PR #17, opened 2021-01-16
Fix: ignoring attributes on template argument

Jamaika1:
```
CommonDef.h:600:55: warning: ignoring attributes on template argument 'void(int, const char*, va_list)' {aka 'void(int, const char*, char*)'} [-Wignored-attributes]
  600 | extern std::function<void( int, const char*, va_list )> g_msgFnc;
      |                                                       ^
```


adamjw24:
What compiler is this?

Jamaika1:
GCC 10.2 or 11.0 C++11

adamjw24:
We cannot reproduce the problem. Also, we are working on cleaning up the lib interface, so maybe the problems will be resolved within this refactoring.

PR #24, opened 2021-03-05
Changed profile description

Jamaika1:


adamjw24:
The mentioned profiles are not supported and should not be shown in help.

See the following snippet:
https://github.com/fraunhoferhhi/vvenc/blob/0f0f08f305f5d1f341c699b27791cae8b83e1d66/source/Lib/vvenc/vvencCfg.cpp#L1392-L1398

PR #31, opened 2021-03-22
Prepend apputils to vvenc_ADD_SUBDIRECTORIES

bart-kroon:
Sorry for bothering you again, but when building VVenC v0.3.1.0 as a sub-project like this:

```cmake
fetchcontent_declare(VVENC
  GIT_REPOSITORY "https://github.com/fraunhoferhhi/vvenc.git"
  GIT_TAG "v0.3.1.0"
  GIT_PROGRESS TRUE)
set(vvenc_ADD_SUBDIRECTORIES "source/App/vvencFFapp")
fetchcontent_makeavailable(VVENC)
```

I get a linker error that `-lapputils` is missing. This pull request fixes that.

If in your opinion this should be part of the client code, then I could change the second-last line in above snippet to explicitly ask for `source/Lib/apputils`, but I just want to build and use `vvencFFapp`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,7 @@ if( NOT CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
   endif()
 
   # vvenc embedded by superproject, always include source/Lib/vvenc  as first component
-  list( PREPEND ${PROJECT_NAME}_ADD_SUBDIRECTORIES "source/Lib/vvenc" )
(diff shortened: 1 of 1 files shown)


adamjw24:
Dear Bart,

thanks for pointing that out. We discussed and think that apputils should not always be added to the build project, because the main use-case is to get the lib, not the app. We found a solution to add apputils conditionally. We will add it to the next version or possibly just push it to the head soon.

I will close the pull request. For now as a workaround apputils can be just specified for inclusion from the outside.

Best regards,
Adam

adamjw24:
See #32

bart-kroon:
> because the main use-case is to get the lib, not the app.

Make sense to me.

PR #38, opened 2021-04-01
Add Github Actions to build static versions

cdgriffith:
Here are three very basic scripts that will build the static releases, as outlined in the readme.  

They will each run the `vvenclibtest` to make sure it compiled correctly and then zip up the `vvencapp`, `vvencFFapp` and `LICNESE.txt`.

Example runs at https://github.com/cdgriffith/vvenc/actions

They currently do NOT include version number in the zip name, nor try to automatically attach to release.

### DIFF
--- a/.github/workflows/build_linux.yaml
+++ b/.github/workflows/build_linux.yaml
@@ -1,0 +1,27 @@
+name: Build executables for Linux
+on:
+  push:
+    branches: [ master, develop, build ]
+  pull_request:
+    branches: [ master, develop ]
+
+jobs:
+  build:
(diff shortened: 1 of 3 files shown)


lvaschmidt:
This seems good to me, do we have a reason this hasn't been merged?

adamjw24:
I'm not sure this is the place to release the binaries for the project. The decision has not yet been made, thats why this MR is still pending.

adamjw24:
We will not be including binaries in the repo for now. You are free to publish them in your fork, though.

Issue #37, opened 2021-03-26
CMake dependency is higher than 3.12 as listed in the README


lvaschmidt:
I tried compiling on CentOS running and preparing with CMake 3.12, which didn't work, but then when I ran the exact same command with CMake 3.20, it compiled fine. 

If I have time later I'll try to pin down which version of CMake is actually required.

jbrdbg:
CMake 3.16 works also fine. We will investigate this issue and provide a fix in a future release.

brixium:
I compiled it successfully in Debian 10, which ships CMake version 3.13.4.

lvaschmidt:
That seems consistent with what I would expect, since I read somewhere that CMake 3.13 added a new place it could look for targets.

When I compile on CMake 3.13.0

```
[root@server vvenc]# /root/cmakevs/cmake-3.13.0/bin/cmake --version
cmake version 3.13.0

CMake suite maintained and supported by Kitware (kitware.com/cmake).
[root@server vvenc]# /root/cmakevs/cmake-3.13.0/bin/cmake .
-- The C compiler identification is GNU 7.3.1
-- The CXX compiler identification is GNU 7.3.1
-- Check for working C compiler: /opt/rh/devtoolset-7/root/usr/bin/cc
-- Check for working C compiler: /opt/rh/devtoolset-7/root/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /opt/rh/devtoolset-7/root/usr/bin/c++
-- Check for working CXX compiler: /opt/rh/devtoolset-7/root/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- CMAKE_MODULE_PATH: updating module path to: /root/vvc/vvenc/cmake/modules
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
-- Configuring done
-- Generating done
-- Build files have been written to: /root/vvc/vvenc
```

When I try to compile with CMake 3.12.4

[...]

PR #44, opened 2021-04-07
Update CMakeLists.txt to require 3.13.0

lvaschmidt:
Bump CMake requirement from 3.12.0 to 3.13.0 in the CMakeLists.txt file to automatically stop someone from encountering issue #37 (As we've established that compiling with 3.12 will fail.)


adamjw24:
makes sense

PR #48, opened 2021-04-08
Remove check for empty input to fix stdin piping

### LINKED ISSUE #47: YUV piping doesn't work
Example of command:
```
ffmpeg -hide_banner -i test.mkv -c:v rawvideo -pix_fmt  yuv420p10le - | ./vvencapp -p 1 --preset faster -o try.h266 --format yuv420_10 --size 1920x1080 --framerate 30 --qp 30 --qpa 0 --input "" -f 30
```
Output:
```
started @ Wed Apr  7 20:37:29 2021

vvencapp [error]: failed to open input file
```
related to this commit https://github.com/fraunhoferhhi/vvenc/commit/b465e4a148ace45fca007f9b927215fc4856ee49

lvaschmidt:
This check always failed, seems to encode fine without it though. Should fix #47

### DIFF
--- a/source/Lib/apputils/YuvFileIO.cpp
+++ b/source/Lib/apputils/YuvFileIO.cpp
@@ -512,15 +512,7 @@ int YuvFileIO::open( const std::string &fileName, bool bWriteMode, const int fil
   {
     if( fileName.empty() )
     {
-      if( ( fseek(stdin, 0, SEEK_END), ftell(stdin)) > 0 )
-      {
-        rewind( stdin );
-      }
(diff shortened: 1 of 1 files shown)


adamjw24:
While this might fix the issue #47, just removing the check is not enough. For now we had "implicit" signaling of the pipe input. The implicit condition was than confirmed by data being available on stdin. If we remove the check, it's better to go for explicit signaling. We are fixing this feature to use the typical input pipe semantics with "-i -" and will push it early next week.

PR #86, opened 2021-08-20
Update README.md

bradh:
Make patent non-licensing explicit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,8 @@ Feel free to contribute. To do so:
 # License
 
 Please see [LICENSE.txt](./LICENSE.txt) file for the terms of use of the contents of this repository.
+    
+In particular, note that the license does not include patent licensing that may be applicable to the use of the contents of this repository.
 
 For more information, please contact: vvc@hhi.fraunhofer.de
 


adamjw24:
Thanks for this proposal. Every wording regarding to the license has to be signed of by way more people than functional changes, so the consideration of this might take some time.

The information is contained in the license file, so I guess it's not a time critical change, but rather a small connivance.

bradh:
Would it be easier if it used the exact wording from the license file?
```
No patent rights, trademark rights and/or 
other Intellectual Property Rights other than the copyrights concerning 
the Software are granted under this license.
```
?

adamjw24:
I'll know in a week at earliest, bcs of vacation and stuff.

Myself, I don't see a point bringing up specific parts of the license for people who don't want to click the link. I looked at that part of the readme, and it doesn't say anything about the license, it just states the copyright and says too look up the license file. If there were other points being discussed in the readme I'd be more agreeable, but the readme doesnt say "in short, the license is..., and for details look up the actual license", it just says "go look up the actual license". If we start doing excerpts of the license, it becomes a whole new discussion, what points should be contained within the excerpt, and what legal meaning does the excerpt have and so on. Do you see the point?

I think the license itself is straightforward enough as it is, and it is what it is. We'll discuss it eventally tho.

bradh:
I think this case is a bit different. Most projects disclaim patent licences,  but actually having a patent pool organised is rare.

adamjw24:
I had a look at x265 and aomenc and neither mentions patent situation in the readme. I really don't think this is the single most important aspect of the license to be mentioned.

PR #90, opened 2021-08-24
add CheckIPOSupported to avoid LTO not supported by compilers

hizia320:
Not all compilers support LTO


adamjw24:
What compilers do not support LTO? We test a wide range of version for GCC/Clang/MSVC, every now and than someone builds it using ICC.

Could you fix the formatting and rename the "result" variable to something meaningful (HAS_LTO/LTO_AVAILABLE or similar)? In the formatting, add a space within the parentheses before first and after last argument. Also remove the additional empty line at the beginning of the changed block.

adamjw24:
Since this is a user setting, maybe add `check_ipo_supported()` into the following block to generate an error if the property is set and not available? https://github.com/fraunhoferhhi/vvenc/blob/0e6ca4febb21facec606fe803273fed31beec6fc/CMakeLists.txt#L195-L199

PR #109, opened 2021-11-15
Remove global variables for the logging callback

ChristianFeldmann:
Using many encoder instances at the same time causes problems if the logging callback uses global variables. 
This is my first attempt at setting a callback to each encoder instance. 
The problems still are: 
 - The decoder lib functions don't support the setting. Why are they even in this repo? This is the encoder and not a decoder? As far as I can see they are never actually used.
 - The tracing also does not have support for this yet. But this can be added
 - I don't understand why in `vvencCfg.h` you are exposing multiple undocumented functions on the config. These also use logging so this does not work yet.

### DIFF
--- a/include/vvenc/vvenc.h
+++ b/include/vvenc/vvenc.h
@@ -379,13 +379,14 @@ VVENC_DECL int vvenc_print_summary( vvencEncoder * );
 VVENC_DECL const char* vvenc_get_error_msg( int nRet );
 
 /* vvenc_set_logging_callback
- This method registers a log message callback function to the encoder library.
+ This method registers a log message callback function to the encoder instance.
  If no such function has been registered, the library will omit all messages.
+ \param[in]  vvencEncoder pointer to opaque handler
  \param[in]  ctx pointer of the caller, if not needed set it to null
  \paramin]   Log message callback function.
  \retval     int VVENC_ERR_INITIALIZE indicates the encoder was not successfully initialized in advance, otherwise the return value VVENC_OK indicates success.
 */
-VVENC_DECL int vvenc_set_logging_callback( void * ctx, vvencLoggingCallback callback );
+VVENC_DECL int vvenc_set_logging_callback( vvencEncoder * , void * ctx, vvencLoggingCallback callback );
 
 /* vvenc_get_compile_info_string
  creates compile info string containing OS, Compiler and Bit-depth (e.g. 32 or 64 bit).
--- a/source/App/vvencFFapp/EncApp.cpp
+++ b/source/App/vvencFFapp/EncApp.cpp
@@ -125,6 +125,8 @@ int EncApp::encode()
     return -1;
   }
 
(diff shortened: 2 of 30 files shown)


adamjw24:
About your comments.

- Decoder module can be used to bootstrap the encoder to a specific state (using `DebugBitstream` or `DebugBitstream1/2` options). This can be useful for debugging/QP switching for fix QP bitrate matching/error resiliency. This is an expert option, probably of no use to everyday users. So far there was always more reason to keep this functionality than kill it for the sake of cleanness.
- The tracing is also an expert option and has different requirements than the rest of the code. It is not thread-safe and thus can use standard IO. This is why it is disabled in standard builds.
- What functionality do you mean in vvencCfg.h? If there is no documentation, the options should probably not be used externally. We might write it down somewhere as a high-level guideline.

About the MR. I don't like using the Slice (which is there to represent slice properties) object to pass the logger around. I understand that the logger has to be universally available and bcs of our program structure there is no top-level "encoding context" object that could store it. How about putting it into vvencCfg object as a member? We will discuss it this week.

ChristianFeldmann:
Thanks for the info. Then we have to see how the decoder and analyzer can do their logging. Or if they are just used for a very specific application / debugging maybe it would be ok if they just log to `std::cout`. 
In the `vvencCfg.h` I mean the functions:
```
VVENC_DECL void vvenc_config_default( vvenc_config *cfg );
VVENC_DECL int vvenc_init_preset( vvenc_config *cfg, vvencPresetMode preset );
VVENC_DECL int vvenc_init_default( vvenc_config *cfg, int width, int height, int framerate, int targetbitrate, int qp, vvencPresetMode preset );
VVENC_DECL bool vvenc_init_config_parameter( vvenc_config *cfg );
VVENC_DECL const char* vvenc_get_config_as_string( vvenc_config *cfg, vvencMsgLevel eMsgLevel );
```
They seem to be exported with `VVENC_DECL` but they are not really documented. 

I agree that the Slice object is not optimal for this. I was also thinking to put it into the config object but then I thought that it is not really a config. Although for other encoders the logging callback is a parameter in the config that you can set. So yes maybe this is an option.

adamjw24:
So we've talked it a bit, and I'd have to hold this back. We are now working on two major features. First, a string-based interface (for better API/ABI consistency in future versions). Second, a new rate control mode requiring two concurrent encoders. Both of those features will have a lot of refactoring and accepting your changes right now would be more of an obstacle, also given how the concept is not yet fully developed.

We will take care of better documentation while extending/refactoring the interface.

About where to put the logger function pointer, I think vvenc_config is not such a bad idea. I mean, in the end, it is a configuration - just a functional one in that case.

If we don't fix the issue in the process of the refactoring work, you're welcome to retackle it after the changes are made. I'd hold of on bigger stuff until than. Closing for now.

ChristianFeldmann:
Thanks for the update. I guess we can live for now without the logging callback. Although in a real production environment I would feel much much better if this works in a multithreading save way.

PR #143, opened 2022-02-21
bugfix: if isLast and DMVR in filterCopy()

hizia320:
fix shifting,
  if isLast==true (and isFirst==false) and DMVR mode
    if IF_INTERNAL_PREC_BILINEAR > clpRng.bd
       => rightShift
    else
       => leftShift

### DIFF
--- a/source/Lib/CommonLib/InterpolationFilter.cpp
+++ b/source/Lib/CommonLib/InterpolationFilter.cpp
@@ -323,12 +323,11 @@ void InterpolationFilter::filterCopy( const ClpRng& clpRng, const Pel* src, int 
       if ((clpRng.bd - IF_INTERNAL_PREC_BILINEAR) > 0)
       {
         shift10BitOut = (clpRng.bd - IF_INTERNAL_PREC_BILINEAR);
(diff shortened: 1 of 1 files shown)


adamjw24:
So the first part somehow doesn't matter, since VVenC cannot do >10bit. For the second part you're saying there is a bug when doing 8-bit encoding without SIMD? We'll have a look and merge if confirmed.

adamjw24:
This code needs some clean, but your fix is not required. Basically the code will never be active.

The `filterCopy` will only be used when `xFrac=yFrac=0`. In that case `isFirst` will always be `true`, since it is the first "filtering" step, and the execution never goes into `isLast && !isFirst` case. I will clean it up a bit and add checks for cases that should not be happening. Closing as irrelevant.

adamjw24:
Also, DMVR bilinear filter implies bi-prediction, which implies `!isLast`

hizia320:
Thanks for your confirmation.

Actually, I just utilized this code to do something and found its logic was not expected.  Of course, I hope it's correct no matter it is active or not in VVenC.

Thanks for your help.

PR #177, opened 2022-07-03
Update CMake link to HTTPS & the redirect target

Schweinepriester:
Pretty much the title: http://www.cmake.org/ is being redirected to https://cmake.org/, so use HTTPS right away and skip the redirect.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -12,4 +12,3 @@
 * Christian Stoffers, , Fraunhofer HHI
 * Gabriel Hege, , Fraunhofer HHI
 * Jens Güther, , Fraunhofer HHI
-* Kai Hollberg, @Schweinepriester,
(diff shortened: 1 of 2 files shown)


adamjw24:
Thank you for looking that up. I hope you don't mind, but I would rather not add you as author with such a minor change. The authors are implicitely, because of our license, copyright holders of the software, which has far-reaching implications. We did it so people don't have to give up copyright when contributing, but I feel like that change is not big enough to extend the group of copyright holders. So if you could not incldue that change I would approve the PR. Otherwise, just from the project organization point of view and for simplicity of the governance, I would rather reject it.

Schweinepriester:
Totally fair, I for sure hesitated adding myself for this change, but since it's plain in the README and I had another project ask me to add me even for such a little change, I've gone ahead initially - will remove it now!

I might suggest changing the corresponding line to something like

> Add your name to [AUTHORS.md](https://github.com/fraunhoferhhi/vvenc/blob/master/AUTHORS.md), if your changes are substantial

though 0:-)

PR #187, opened 2022-08-27
Delete duplicate 'inline' for AVX512

Jamaika1:
```
x86/CommonDefX86.h:331:15: error: duplicate 'inline'
  331 | ALWAYS_INLINE inline __m512i
      |               ^~~~~~
```


adamjw24:
Could you please rather remove the `ALWAYS_INLINE` than `inline`?

Best regards,
Adam

adamjw24:
Actually, it should `static inline`. Could you also fix that while you're at it?

PR #237, opened 2023-02-02
bugfix, xCodeRefPicList()

hizia320:
st_ref_pic_flag coded is dependent on sps_long_term_ref_pics_flag, not numberOfLongtermPictures


adamjw24:
Looks good too me, but will need to consult with the spec. Will probably merge next week.

hizia320:
compared with VTM, it should be as fixed

adamjw24:
Yeah, looks good.

PR #243, opened 2023-02-23
refactor: Remove direct copy of nlohmann/json

howjmay:
In this way we can make the porject smaller and easier to update


adamjw24:
This seems to break the build. Any idea why that is? How do you pull the submodule?

adamjw24:
We talked about that and it this seems like we'd rather keep a copy of the repo rather than use git to pull it. This way a clone of our repo is always self contained and can be directly built, even in private networks without access to internet, as some of our CI is.

howjmay:
got it thanks!!

PR #267, opened 2023-04-11
cmake: Fix compiling if CFLAGS or CXXFLAGS are set

Biswa96:
This fixes checking _mm_extract_epi64 intrinsic when CFLAGS or CXXFLAGS
    are set. The -msse4.1 option is appended with existing compiler flags.
    Otherwise, _mm_extract_epi64 check fails without the -msse4.1 option.
    For example, the issue has been seen with the following variables in
    mingw-w64 gcc 64-bit toolchain.

    export CPPFLAGS="-D_FORTIFY_SOURCE=0 -D__USE_MINGW_ANSI_STDIO=1"
    export CFLAGS="-mthreads -mtune=generic -O2 -pipe"
    export CXXFLAGS="-mthreads -mtune=generic -O2 -pipe"
    export LDFLAGS="-pipe -static-libgcc -static-libstdc++"

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,8 @@ if( VVENC_ENABLE_X86_SIMD )
     # when building for non-x86, but emulating simd using simd-everywhere (e.g. on ARM),
     # the x86-compiler flags are not understood by the compiler
     set_if_compiler_supports_flag( FLAG_msse41 -msse4.1 )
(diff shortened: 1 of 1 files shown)


adamjw24:
Could you elaborate more? Why does `add_compile_options` ignore your system-wide `CPPFLAGS` and `CFLAGS` settings?

adamjw24:
Obsoleted by #268

PR #268, opened 2023-04-11
Proper fix for sse 41 detection

adamjw24:
`_mm_extract_epi64` is a SSE4.1 intrinsic, so the check needs to use an appropriate flag


Biswa96:
I have verified that this change also fixes the build issue with mingw gcc and clang. Thank you.

adamjw24:
> I have verified that this change also fixes the build issue with mingw gcc and clang. Thank you.

Thanks for reporting

PR #334, opened 2023-11-24
Fix slice chroma QP adjust MinCuSize error.

luofalei:
Fix a typo.

### DIFF
--- a/source/Lib/EncoderLib/EncCu.cpp
+++ b/source/Lib/EncoderLib/EncCu.cpp
@@ -666,15 +666,13 @@ void EncCu::xCompressCU( CodingStructure*& tempCS, CodingStructure*& bestCS, Par
 
   m_cuChromaQpOffsetIdxPlus1 = 0;
 
-  if( slice.chromaQpAdjEnabled )
+  if( slice.chromaQpAdjEnabled && partitioner.currQgChromaEnable() )
   {
     // TODO M0133 : double check encoder decisions with respect to chroma QG detection and actual encode
+    int cuChromaQpOffsetSubdiv = slice.isIntra() ? slice.picHeader->cuChromaQpOffsetSubdivIntra : slice.picHeader->cuChromaQpOffsetSubdivInter;
     int lgMinCuSize = sps.log2MinCodingBlockSize +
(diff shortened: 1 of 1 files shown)


source/Lib/EncoderLib/EncCu.cpp
```diff
@@ -670,7 +670,7 @@ void EncCu::xCompressCU( CodingStructure*& tempCS, CodingStructure*& bestCS, Par
   {
     // TODO M0133 : double check encoder decisions with respect to chroma QG detection and actual encode
     int lgMinCuSize = sps.log2MinCodingBlockSize +
-      std::max<int>(0, floorLog2(sps.CTUSize - sps.log2MinCodingBlockSize - int((slice.isIntra() ? slice.picHeader->cuChromaQpOffsetSubdivIntra : slice.picHeader->cuChromaQpOffsetSubdivInter) / 2)));
+      std::max<int>(0, floorLog2(sps.CTUSize) - sps.log2MinCodingBlockSize - int((slice.isIntra() ? slice.picHeader->cuChromaQpOffsetSubdivIntra : slice.picHeader->cuChromaQpOffsetSubdivInter) / 2));
```
adamjw24, line 673: Looks good to me, good catch!

Could you move the assignment into the `partitioner.currQgChromaEnable()` condition? Than the two conditions could be combined together. Btw, what is the reasoning behind the assignment. We inherited it from VTM and it does not look like it makes sense to me...

adamjw24: I traced this code way back to HM, and its indeed just for testing, but makes no sense for production. The code should still be braced correctly, tho.

If you could refactor the two ifs into one, it would be amazing.

luofalei: Thanks for your suggestions, please check the updated commit. I found this issue when trying to optimize block-level QPs for better rate allocation.

adamjw24: Thanks. I'll pull the PR after the pipeline has finished. 

I traced the code back to HM: https://vcgit.hhi.fraunhofer.de/jvet/HM/-/blob/master/source/Lib/TLibEncoder/TEncCu.cpp?ref_type=heads#L685

There it is explained that this is only exemplary to confirm lookup in the PPS list works fine. So you would need your own algorithm for the allocation.

Let me know if you find an interesting approach, and are interested in putting it into VVenC.

luofalei: Updated to align VTM,  add +1 before / 2.

Issue #410, opened 2024-07-31
VVENC possibly overruns the stack with libc-musl


AlexanderSchuetz97:
I am compiling a static build of FFMPEG with vvenc on alpine linux. Alpine linux uses libc-musl.
A notable difference between libc-musl and glibc is that the default stack size is significantly smaller.
To test my theory I added the following to the linker arguments and recompiled ffmpeg and vvenc:
```
export LDFLAGS=-Wl,-z,stack-size=1048576
```
and then it no longer crashes and works as expected.

The call to ffmpeg that uses vvenc is:
```
/tmp/ffmpeg -y -v error -video_size 1280x720 -f rawvideo -pix_fmt bgr24 -i unix:///tmp/17438466013925160049.sock -f mp4 -vcodec vvc -r 30/1 -an /tmp/output16127279067242218599.bin
```
Its unlikely that the rawvideo decoder is the culprit here. Especially since if I use another codec like mpeg1video with the same call
it does not crash. This leaves only vvenc as the possible culprit.

Does vvenc create threads where pthread_attr_setstacksize is not called? 
In this case libc musl allocates only between 80 and 128k for the stack. 
This may not be enough for larger allocations.

AlexanderSchuetz97:
Well trying to get vvenc to work at all on alpine linux (presumeably with or without ffmpeg) requires setting the LDFLAGS as mentioned. The problem is that this reserves 10MB of stack for all threads in all of ffmpeg not just the threads that actually need that much stack. 10MB is coincidentally the amount that most versions of glibc choose for the stack size. I dont think you need that much stack, but since compiling ffmpeg takes a significant timea and I do not quite know what determines your alloca's size I decided to stop experimenting and report this here.

libkvazaar (h265 encoder) had the same issue as vvenc. 
They solved it by calling pthread_attr_setstacksize with a size large enough for their alloca's. This is portable across all libc implementations as far as I know, so it should work with glibc/libc-musl and whatever else there is out there.

As mentioned the workaround with LDFLAGS works but this is not that "clean". I would prefer if the application set the required stack size to a reasonable amount and doesn't rely on the default behavior of the c-library. This would make vvenc more portable so its a good idea to do this regardless.

Since I have a workaround I am under no time pressure.

adamjw24:
Thanks, we'll look into the `pthread_attr_setstacksize` function to ensure our code works independent of the underlying libc implementation.

K-os:
[...]

PR #411, opened 2024-08-07
pthreads specific thread implementation to set stack size

K-os:
Adds a thread class implemented with pthreads to set the stack size to at least 1MB. It exposes an interface similar to std::thread, so it is interchangeable with that for our use case.

fixes GitHub issue [#410](https://github.com/fraunhoferhhi/vvenc/issues/410)

### DIFF
--- a/source/Lib/Utilities/NoMallocThreadPool.cpp
+++ b/source/Lib/Utilities/NoMallocThreadPool.cpp
@@ -48,10 +48,6 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "NoMallocThreadPool.h"
 
 
-#if __linux
-#include <pthread.h>
-#endif
-
 //! \ingroup Utilities
 //! \{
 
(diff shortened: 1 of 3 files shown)


source/Lib/Utilities/NoMallocThreadPool.cpp
```diff
@@ -270,6 +274,77 @@ bool NoMallocThreadPool::processTask( int threadId, NoMallocThreadPool::Slot& ta
...
+  {
+    std::unique_ptr<WrappedCall> call( static_cast<WrappedCall*>( p ) );
+
+    ( *call )();
+
+    return nullptr;
+  };
+
+  pthread_attr_t attr;
+  CHECK( pthread_attr_init( &attr ) != 0, "pthread_attr_init() failed" );
```
adamjw24, line 296: please don't put actual functionality into CHECK()-Macro. Those might be stripped down eventually. Only perform checks there, don't put actual code there.

PR #420, opened 2024-09-25
Avoid checking for arm/aarch64 in toolchain path

georges-arm:
Since the `CMAKE_CXX_COMPILER` contains the full path including directory components, this check will enable Arm intrinsics even on non-Arm platforms if the directory the compiler resides in happens to contain the substring "arm", for example:

        /home/parma/toolchains/g++-14
               ^^^

`CMAKE_SYSTEM_PROCESSOR` already specifies the target architecture so there is no need to additionally check the compiler string, therefore we simply remove this condition.


adamjw24:
LOL, good one.

Thanks for the fix, we'll have a look and probably merge soon.

K-os (approved):

georges-arm:
Thanks for the quick review!

> Have you tried building VVdeC with your setup? Did that work? In VVdeC the heuristics are already as they will be in VVenC after the next update. Some feedback would be nice.

I just had a quick try and VVdeC seems to work fine on both x86_64 and AArch64 as well as for cross compiling. 👍

PR #421, opened 2024-09-25
Fix build failures on AArch64 with GCC-14

georges-arm:
When building with GCC-14 on an AArch64 machine, the build currently fails with `-Werror` due to a couple of warnings that are turned into errors:

* GCC appears to not understand the indexing for the `coeffV` array load and incorrectly reports this as an out of bounds access. Add a new `GCC_WARNING_DISABLE_uninitialized` to handle this.

* GCC reports the ternary setting `f` in `IntraPredAngleLumaCore_SIMD` as being set from a maybe-uninitialised source. Splitting the ternary appears to be sufficient to suppress this.

### DIFF
--- a/source/Lib/CommonLib/CommonDef.h
+++ b/source/Lib/CommonLib/CommonDef.h
@@ -96,6 +96,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #if defined( __GNUC__ ) && __GNUC__ >= 8 && !defined( __clang__ )
 # define GCC_WARNING_DISABLE_maybe_uninitialized _Pragma("GCC diagnostic push"); \
                                                  _Pragma("GCC diagnostic ignored \"-Wmaybe-uninitialized\"");
+# define GCC_WARNING_DISABLE_uninitialized       _Pragma("GCC diagnostic push"); \
+                                                 _Pragma("GCC diagnostic ignored \"-Wuninitialized\"");
 # define GCC_WARNING_DISABLE_class_memaccess     _Pragma("GCC diagnostic push"); \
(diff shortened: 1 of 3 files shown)


adamjw24:
This is with which exact version of GCC? Sounds like GCC-14.0 early version issues that would've been fixed in 14.1 or 14.2.

georges-arm:
> This is with which exact version of GCC? Sounds like GCC-14.0 early version issues that would've been fixed in 14.1 or 14.2.

Indeed it's a pretty early version of GCC:

```
$ gcc-14 --version
gcc-14 (Ubuntu 14-20240412-0ubuntu1) 14.0.1 20240412 (experimental) [master r14-9935-g67e1433a94f]
Copyright (C) 2024 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

But this is the version that got installed when I did `apt install gcc-14` on an Ubuntu 24.04 machine so not like it was my own hacked together version of GCC :)

This is only particularly problematic for VVenC since it unconditionally enables `-Werror`. To enable building with in-development compilers (or compilers that get released in the future that don't even exist yet) that may enable new warnings that don't appear in today's compilers it is more common for software to provide a separate option for enabling `-Werror` and then have that enabled in CI or as a pre-commit check instead, that way customers who are only interested in building and consuming the library don't get caught out by such things.

adamjw24:
So... I am pretty sure those are some false positive by an over-eager compiler. Please see #422, which allows you to build even with compiler warning (disabled by default). I think this is a preferable way to developing workarounds for buggy compilers.

Thanks for pointing that out and preparing a fix anyway.

PR #426, opened 2024-10-02
Add CONFIGURE_DEPENDS to CMake GLOB lists

georges-arm:
The addition of new source files will not ordinarily trigger them to be added to the build due to how `GLOB`s are implemented in CMake.

CMake 3.12 added a new `CONFIGURE_DEPENDS` flag to enable CMake to re-check `GLOB` results during build time and re-generate the build scripts if necessary. The existing `CMakeLists.txt` already specifies a minimum CMake version of 3.13 so this should always be available.

### DIFF
--- a/source/App/vvencFFapp/CMakeLists.txt
+++ b/source/App/vvencFFapp/CMakeLists.txt
@@ -2,10 +2,10 @@
 set( EXE_NAME vvencFFapp )
 
 # get source files
-file( GLOB SRC_FILES "*.cpp" )
+file( GLOB SRC_FILES CONFIGURE_DEPENDS "*.cpp" )
 
 # get include files
-file( GLOB INC_FILES "*.h" )
+file( GLOB INC_FILES CONFIGURE_DEPENDS "*.h" )
 
 # get address sanitizer libs for gcc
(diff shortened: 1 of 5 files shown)


K-os:
I would suggest adding `CONFIGURE_DEPENDS` to the GLOBs in the other CMakeLists.txt files also:
- source/App/vvencFFapp/CMakeLists.txt
- source/App/vvencapp/CMakeLists.txt
- test/vvenclibtest/CMakeLists.txt
- test/vvencinterfacetest/CMakeLists.txt

Besides that, it looks good to me.

georges-arm:
> I would suggest adding `CONFIGURE_DEPENDS` to the GLOBs in the other CMakeLists.txt files also:
> 
> * source/App/vvencFFapp/CMakeLists.txt
> * source/App/vvencapp/CMakeLists.txt
> * test/vvenclibtest/CMakeLists.txt
> * test/vvencinterfacetest/CMakeLists.txt
> 
> Besides that, it looks good to me.

Ack! Will amend over the next couple of days to include those files. Thanks for the suggestion.

K-os (approved):
Thanks for the fix. Looks good now.

PR #431, opened 2024-10-09
Refactor AArch64 Interpolation Filter 16x16 implementation

georges-arm:
The existing `InterpolationFilterARM.h` is only used in one place and would not share much code with an eventual SVE implementation, so simply merge it into the existing `InterpolationFilter_neon.cpp` file. This mirrors what already exists for MCTF on AArch64.

* Also add missing guard in the `.cpp` file such that the code is now guarded by `ENABLE_SIMD_OPT_MCIF`.

The existing `simdFilter16xX_N8_neon` implementation is improved in a few ways:

* The use of a `vsrcv` temporary array rather than simple local variables meant that compilers emit an unnecessary number of load/store instructions in the inner loops. Refactoring this to make the dependency between loop iterations more explicit allows for much nicer generated code.

* The existing code makes use of horizontal reduction (`vaddvq_s32`) instructions for combining the intermediate multiplication products from the filter down to a single value, however we can instead make use of a series of pairwise additions (`vpaddq_s32`) to perform the reduction of multiple vectors at once while maintaining the results in vector registers rather than scalar registers.

* In the `!isLast` case we can take advantage of the final shift amount being a known constant to use an instruction taking an immediate.

Additionally this refactoring allows us to remove the `GCC_WARNING_DISABLE_maybe_uninitialized` macro usage from this file since the warning no longer occurs.

Running a video encoding job on a Neoverse V2 machine using the `--preset=fast` setting shows a ~1.8% improvement in reported FPS.


adamjw24:
Nice, thanks!

The guy who "owns" this code will be back in a week or so and I'll ask him to review then.

georges-arm:
> Looks good, got positive review on this.

Good to hear!

> Just one question, some of the stuff is also used in VVdeC. Would it be ok for us to port your improvements there as well? If so, any requirements? It would be ok to add you as author there, even if its us moving your code. So let us know. I mean this stuff and any potential future improvements as well.

That's fine, if anything that saves me from needing to eventually port it myself. Thanks!

PR #436, opened 2024-10-14
Fix added integer int

Jamaika1:
```
TrQuant.cpp: In member function 'void vvenc::TrQuant::xIT(const vvenc::TransformUnit&, vvenc::ComponentID, const CCoeffBuf&, vvenc::PelBuf&)':
TrQuant.cpp:624:33: warning: comparison of unsigned expression in '< 0' is always false [-Wtype-limits]
  624 |     CHECK((transformHeightIndex < 0), "There is a problem with the height.");
      |            ~~~~~~~~~~~~~~~~~~~~~^~~
TypeDef.h:641:32: note: in definition of macro 'CHECK'
  641 | #define CHECK(c,x)          if(c){ THROW(x); }
      |                                ^
```


adamjw24:
It seems a more appropriate way of fixing this would be to delete the CHECK

Jamaika1:
I understand that transformHeightIndex is never negative.
I delete negative value of transformHeightIndex.

PR #437, opened 2024-10-14
Fix added this local variables by value

Jamaika1:
```
NoMallocThreadPool.h: In lambda function:
NoMallocThreadPool.h:220:31: warning: implicit capture of 'this' via '[=]' is deprecated in C++20 [-Wdeprecated]
  220 |     nonconst->m_cond.wait( l, [=] { return m_count == 0; } );
      |                               ^
NoMallocThreadPool.h:220:31: note: add explicit 'this' or '*this' capture
```

### DIFF
--- a/source/Lib/Utilities/NoMallocThreadPool.h
+++ b/source/Lib/Utilities/NoMallocThreadPool.h
@@ -157,7 +157,7 @@ struct BlockingBarrier
     BlockingBarrier* nonconst = const_cast<BlockingBarrier*>(this);
 
     std::unique_lock<std::mutex> l( nonconst->m_lock );
-#if __GNUC__ && (__cplusplus > 201703L)
+#if __MINGW32__ && (__cplusplus > 201703L)
     nonconst->m_cond.wait( l, [=, this] { return !m_intBarrier.isBlocked(); } );
 #else
     nonconst->m_cond.wait( l, [=] { return !m_intBarrier.isBlocked(); } );
@@ -214,7 +214,7 @@ struct WaitCounter
     WaitCounter* nonconst = const_cast<WaitCounter*>(this);
 
     std::unique_lock<std::mutex> l( nonconst->m_lock );
-#if __GNUC__ && (__cplusplus > 201703L)
+#if __MINGW32__ && (__cplusplus > 201703L)
     nonconst->m_cond.wait( l, [=, this] { return m_count == 0; } );
 #else
     nonconst->m_cond.wait( l, [=] { return m_count == 0; } );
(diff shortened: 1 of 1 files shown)


source/Lib/Utilities/NoMallocThreadPool.h
```diff
@@ -157,7 +157,11 @@ struct BlockingBarrier
     BlockingBarrier* nonconst = const_cast<BlockingBarrier*>(this);
 
     std::unique_lock<std::mutex> l( nonconst->m_lock );
+#if __GNUC__ && (__cplusplus > 201703L)
```
adamjw24, line 160: Is there a reason to guard the fix with __GNUC__ and CPP version? Why not just make it unconditional?

Jamaika1: hmm, if vvenc is for c++17 then there is no problem.
Unfortunately I recently added plugin for ffmpeg in gnu++23 and I have warnings.
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0806r2.html

adamjw24: The question is. Does the code compile in c++17? If yes, there is no need to use two version of this code.

adamjw24: As mentioned, please remove the guard and just change the code, unless its not c++17 compliant.

jakubjelinek:
Explicit capture of this is only valid in C++20 and later.  See the C++20 https://wg21.link/p0409r2 paper.

adamjw24:
> Explicit capture of this is only valid in C++20 and later. See the C++20 https://wg21.link/p0409r2 paper.

So there as is a new feature that is syntax incompatible with previous versions but not using it breaks the code as well? Great design philosophy/s.

adamjw24:
@Jamaika1 our code is written using C++14 (`set( CMAKE_CXX_STANDARD 14 )`), so your guard is guaranteed to be false, is thus unnecessary. I don't know why and how this is overwritten, but the fact that it is is your actual bug. Closing.

jwakely:
> > Explicit capture of this is only valid in C++20 and later. See the C++20 https://wg21.link/p0409r2 paper.
> 
> So there as is a new feature that is syntax incompatible with previous versions but not using it breaks the code as well? Great design philosophy/s.

Jakub's comment is a bit misleading.

There's a new feature that is only valid in C++20, i.e. `[=, this]`.

But the previous behaviour of `[=]` can just be replaced with `[this]` in your code, because the **only** thing you're trying to capture is `this`. That is perfectly valid in all versions of C++ from C++11 upwards.


So the correct fix here is to replace `[=]` with `[this]` and not have any `#if` conditional code.

source/Lib/Utilities/NoMallocThreadPool.h
jwakely: There is absolutely no reason to check `__MINGW32__ or `__GNUC__` here. Nothing about the code is GCC-specific.

jwakely:
(and it was only a warning anyway, so the old syntax _is_ still compatible in C++20)

K-os:
@jwakely : Thank you very much for the clarification. I didn't know `[=, this]` was disallowed pre c++17. I thought it was just redundant.

PR #438, opened 2024-10-14
Added typedef va_list

Jamaika1:
```
CommonDef.h:569:62: warning: ignoring attributes on template argument 'void(void*, int, const char*, va_list)' {aka 'void(void*, int, const char*, char*)'} [-Wignored-attributes]
  569 | extern std::function<void( void*, int, const char*, va_list )> g_msgFnc;
      |                                                              ^
```

### DIFF
--- a/source/Lib/CommonLib/CommonDef.h
+++ b/source/Lib/CommonLib/CommonDef.h
@@ -564,6 +564,9 @@ template <typename T> inline void Check3( T minVal, T maxVal, T a)
   CHECK( ( a > maxVal ) || ( a < minVal ), "ERROR: Range check " << minVal << " >= " << a << " <= " << maxVal << " failed" );
 }  ///< general min/max clip
 
+#if __GNUC__ && __cplusplus
+typedef char *  va_list;
+#endif
 // global logger message callback function - DEPRECATED - will be removed in next major version
 extern std::function<void( void*, int, const char*, va_list )> g_msgFnc;
 extern void * g_msgFncCtx;
--- a/source/Lib/CommonLib/CommonDef.h
+++ b/source/Lib/CommonLib/CommonDef.h
@@ -564,7 +564,7 @@ template <typename T> inline void Check3( T minVal, T maxVal, T a)
   CHECK( ( a > maxVal ) || ( a < minVal ), "ERROR: Range check " << minVal << " >= " << a << " <= " << maxVal << " failed" );
 }  ///< general min/max clip
 
-#if __GNUC__ && __cplusplus
+#if __MINGW32__ && __cplusplus
 typedef char *  va_list;
 #endif
 // global logger message callback function - DEPRECATED - will be removed in next major version


source/Lib/CommonLib/CommonDef.h
```diff
@@ -564,6 +564,9 @@ template <typename T> inline void Check3( T minVal, T maxVal, T a)
   CHECK( ( a > maxVal ) || ( a < minVal ), "ERROR: Range check " << minVal << " >= " << a << " <= " << maxVal << " failed" );
 }  ///< general min/max clip
 
+#if __GNUC__ && __cplusplus
+typedef char *  va_list;
+#endif
 // global logger message callback function - DEPRECATED - will be removed in next major version
```
adamjw24, line 570: Looks like this one will going to fixed in the next version anyway.

adamjw24:
I really don't see why this is required. We `#include <cstdarg>` and more should not be required to support this type. Most of all, why would it be required to define that type? I think your compiler is maybe broken?

lehmann-c:
Can you specify you build environment so we can reproduce the issue? (OS, mingw version, build command).
As a fallback, can you test if it builds when you replace the include in file CommonDef.h, by replacing:
`#include <cstdarg>`
to
`#include <stdarg.h>

Jamaika1:
> I really don't see why this is required. We `#include <cstdarg>` and more should not be required to support this type. Most of all, why would it be required to explicitely define that type - it is a C++ standard type after all? I think your compiler is maybe broken?

This is very uncomfortable question. I think that gnu or mingw did not buy the patent, but as they write often not compatible with the standard we do not add like MSVC.


There seems to be a problem with <vadefs.h>

```
#ifndef _VA_LIST_DEFINED	/* if stdargs.h didn't define it */
#define _VA_LIST_DEFINED
#if defined(__GNUC__)
  typedef __gnuc_va_list va_list;
#elif defined(_MSC_VER)
  typedef char *  va_list;
#elif !defined(__WIDL__)
#error VARARGS not implemented for this compiler
#endif
#endif /* _VA_LIST_DEFINED */
```
I think I didn't mistake the link. I'm using gcc version 11.xx and the -O3 option. Higher versions of gcc won't work with this version.
https://github.com/brechtsanders/winlibs_mingw/releases/download/11.5.0posix-12.0.0-ucrt-r1/winlibs-x86_64-posix-seh-gcc-11.5.0-mingw-w64ucrt-12.0.0-r1.zip

lehmann-c:
There is no problem with gcc or mingw as you can see in the [media-autobuild_suite](https://github.com/m-ab-s/media-autobuild_suite). Currently gcc 14.1 with mingw is used to build VVenC and VVdeC for FFmpeg without any problems.

I also tried the [minlibs_mingw  ](https://github.com/brechtsanders/winlibs_mingw/releases/download/11.5.0posix-12.0.0-ucrt-r1/winlibs-x86_64-posix-seh-gcc-11.5.0-mingw-w64ucrt-12.0.0-r1.zip) version from brechtsanders that you was referring to. 
Project was building without any issues. Everything is compiling properly.  
My test build command:  
` cmake -S . -B build/release-static  -G Ninja -DBUILD_SHARED_LIBS=off -DCMAKE_BUILD_TYPE=Release  -DVVENC_ENABLE_LINK_TIME_OPT=OFF -DVVENC_INSTALL_FULLFEATURE_APP=ON -DCMAKE_CXX_FLAGS="-O3 " -DCMAKE_C_FLAGS="-O3  "`
`cmake --build build/release-static -j`  

So if you still have any issues, please provide your full build command with logs so we can help.

Jamaika1:
Thanks for the reply. For those interested, here's my work in 7z files. I don't claim to know everything.
https://github.com/Jamaika1/plugins_ffmpeg701_mingw1150

I had suggestions on the forum why I use -03 and not -02.
https://gcc.gnu.org/bugzilla/buglist.cgi?bug_status=UNCONFIRMED&bug_status=NEW&bug_status=ASSIGNED&bug_status=SUSPENDED&bug_status=WAITING&bug_status=REOPENED&field0-0-0=alias&field0-0-1=short_desc&field0-0-2=status_whiteboard&field0-0-3=content&list_id=447135&order=changeddate%20DESC%2Cpriority%2Cbug_severity&query_format=advanced&type0-0-0=substring&type0-0-1=substring&type0-0-2=substring&type0-0-3=matches&value0-0-0=o3&value0-0-1=o3&value0-0-2=o3&value0-0-3=%22o3%22
... because codecs like aom or vvenc don't work

I stick to my guns. I know that gcc/mingw automatically makes the deprecated information. It was just my suggestion whether va_list shouldn't be separated. I don't see any definition added here that activates va_list.

Latest test O2
https://www.sendspace.com/file/lq6xyt
Latest test O3
in a moment

As for the vvenc ffmpeg plugin then I know it doesn't work for me.
I changed the entry 'params.m_IntraPeriodSec = s->refreshsec'.

lehmann-c:
After having a look into  you github repository, there is no clue how you build your project. Neither any Makefiles nor cmake. 
How shall someone help if you don´t provide your workflow?
This is crucial, particularly as you copied ffmpegs libavcodec files only partly into your project and also changed the files to your workflow.

PR #439, opened 2024-10-14
Fix added integer int

Jamaika1:
```
vvencCfg.cpp:1923:57: warning: comparison of unsigned expression in '< 0' is always false [-Wtype-limits]
 1923 |   vvenc_confirmParameter( c, (c->m_contentLightLevel[0] < 0 || c->m_contentLightLevel[0] > 10000),  "max content light level must 0 <= cll <= 10000 ");
      |                               ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
vvencCfg.cpp:1924:57: warning: comparison of unsigned expression in '< 0' is always false [-Wtype-limits]
 1924 |   vvenc_confirmParameter( c, (c->m_contentLightLevel[1] < 0 || c->m_contentLightLevel[1] > 10000),  "max average content light level must 0 <= cll <= 10000 ");
      |                               ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
```


Jamaika1:
I understand that light level is never negative.
I delete negative value of light level.

PR #442, opened 2024-10-15
cmake: add cache variables for BUILD_SHARED_LIBS & VVENC_LIBRARY_ONLY

### LINKED ISSUE #433: Please add 2 bool cmake entries for cmake-gui
Hi, I think may be add these 2 entries for cmake-gui is better for users build from scratch.

BUILD_SHARED_LIBS: bool, for users who want static or shared builds.

BUILD_PROGRAMS: bool, for users who want library builds only.

K-os:
closes #433

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,6 +103,10 @@ endif()
 # enable install target
 set( VVENC_ENABLE_INSTALL                   ON  CACHE BOOL   "Enable or disable install target" )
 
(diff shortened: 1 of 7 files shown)


lehmann-c (approved):

adamjw24:
What happens to `make test` if you do that?

K-os:
Make test will fail naturally fail if you set VVENC_LIBRARY_ONLY=1, but I guess that is how it is. How should you run the tests, when you don't build the tests?
You can still go to your build directory (e.g. build/release-static), explicitly build all the binaries, and then run ctest.

K-os:
We do output a warning now that the test suite will not be able to run, when enabling VVENC_LIBRARY_ONLY.

PR #453, opened 2024-10-24
Fix compilation for 32-bit Arm platforms

georges-arm:
The existing Neon code makes use of some intrinsics which are only available on AArch64, so adjust these implementations to also work on Armv7 platforms.

Also introduce a new `REAL_TARGET_AARCH64` macro to distinguish between 32-bit and 64-bit in new and existing Arm code.

Since these implementations require that Neon is available, also add `-mfpu=neon` to the compile line for these files. This necessitates removing the SIMD implementation files from LTO in the same way that is already done for the x86 kernels.


source/Lib/vvenc/CMakeLists.txt
```diff
@@ -133,13 +134,20 @@ if( VVENC_ENABLE_ARM_SIMD )
...
-  #                                             INTERPROCEDURAL_OPTIMIZATION_RELEASE        OFF
-  #                                             INTERPROCEDURAL_OPTIMIZATION_RELWITHDEBINFO OFF
-  #                                             INTERPROCEDURAL_OPTIMIZATION_MINSIZEREL     OFF )
+
+  # Disble LTO for the files compiled with special architecture flags.
+  set_target_properties( ${LIB_NAME}_arm_simd PROPERTIES
+                                              INTERPROCEDURAL_OPTIMIZATION                OFF
+                                              INTERPROCEDURAL_OPTIMIZATION_RELEASE        OFF
+                                              INTERPROCEDURAL_OPTIMIZATION_RELWITHDEBINFO OFF
+                                              INTERPROCEDURAL_OPTIMIZATION_MINSIZEREL     OFF )
```
K-os, line 150: Can we enable `-mfpu=neon` for all source files of the whole vvenc library? Then we don't need to  disable LTO for the SIMD vvec_arm_simd target. This was required for x86, because there we detect which SIMD level is available at runtime. But for ARM we always assume NEON is available, so we could enable it for the whole library.

K-os:
Thanks for incorporating my suggestions. From your code around enabling SVE/SVE2 I assume these extensions are not available for ARMv7 ? Then I think the split between `ARM` and `AARCH64` for `VVENC_TARGET_ARCH` really makes sense.

Concerning the interaction between `VVENC_ENABLE_X86_SIMD` and `VVENC_ENABLE_ARM_SIMD`, my intention was that all combinations are possible. I know it worked at some point in time, but I am not sure about the current status. I will look into it, when your changes are merged.

georges-arm:
> Thanks for incorporating my suggestions. From your code around enabling SVE/SVE2 I assume these extensions are not available for ARMv7 ?

Your interpretation is correct. Both SVE and SVE2 are only available in the 64-bit architecture (available from Armv8.2-A and Armv9.0-A respectively).

K-os (approved):
Ok, looks good to me, then.

adamjw24:
Thanks for the MR @georges-arm and for the review @K-os

georges-arm:
> Concerning the interaction between `VVENC_ENABLE_X86_SIMD` and `VVENC_ENABLE_ARM_SIMD`, my intention was that all combinations are possible. I know it worked at some point in time, but I am not sure about the current status. I will look into it, when your changes are merged.

This should be fixed by #455. Thanks for confirming the intent here.

PR #455, opened 2024-10-28
Rework setSIMDExtension to compile with partial SIMD enabled

georges-arm:
There are existing options to enable both Arm native and SIMDe-based intrinsics implementations, however if only one of these is enabled then compilation for Arm targets currently fails.

Fix this by adjusting the existing `#ifdef` guards and adding new ones to cover the previously failing option combinations. Also stop checking `REAL_TARGET_ARM` since it does not distinguish between which SIMD targets are enabled.

Additionally the existing Neon code in RdCostARM.h depends on being able to call into the existing SIMDe-based x86 kernels. If `VVENC_ENABLE_X86_SIMD` is disabled in the CMake then these kernels are unavailable and compilation
fails. Until native Neon kernels are added for the missing functions, simply disable the optimised RdCost Neon code unless the SIMDe kernels are also available.

### DIFF
--- a/source/Lib/CommonLib/arm/RdCostARM.h
+++ b/source/Lib/CommonLib/arm/RdCostARM.h
@@ -51,12 +51,14 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "CommonLib/CommonDef.h"
 #include "../RdCost.h"
 
-#if SIMD_EVERYWHERE_EXTENSION_LEVEL_ID==X86_SIMD_AVX2
-# define USE_AVX2
-#elif SIMD_EVERYWHERE_EXTENSION_LEVEL_ID==X86_SIMD_SSE42
-# define USE_SSE42
-#elif SIMD_EVERYWHERE_EXTENSION_LEVEL_ID==X86_SIMD_SSE41
-# define USE_SSE41
+#if defined( TARGET_SIMD_X86 )
+#if SIMD_EVERYWHERE_EXTENSION_LEVEL_ID == X86_SIMD_AVX2
+#define USE_AVX2
+#elif SIMD_EVERYWHERE_EXTENSION_LEVEL_ID == X86_SIMD_SSE42
+#define USE_SSE42
+#elif SIMD_EVERYWHERE_EXTENSION_LEVEL_ID == X86_SIMD_SSE41
+#define USE_SSE41
+#endif
 #endif
 
 #ifdef TARGET_SIMD_X86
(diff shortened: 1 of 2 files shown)


K-os:
Thank you for attempting to clean up the compilation mess.

I just looked into it and realized that the ARM-SIMD code cannot work without having `TARGET_SIMD_X86` enabled, because it explicitly calls some X86-SIMD implementations (`xCalcHAD*_SSE` for non-square block sizes). This is something that came in after the first NEON implementations, and I did not know this when I said it was supposed to work in my comment on #453. I checked back with colleagues and we can define that `TARGET_SIMD_X86` is a requirement for `TARGET_SIMD_ARM`.

So, even with your changes applied, the code does not compile when `TARGET_SIMD_ARM` is enabled, but `TARGET_SIMD_X86` is not. Maybe we should explicitly catch that in the configure step.

georges-arm:
> Thank you for attempting to clean up the compilation mess.
> 
> I just looked into it and realized that the ARM-SIMD code cannot work without having `TARGET_SIMD_X86` enabled, because it explicitly calls some X86-SIMD implementations (`xCalcHAD*_SSE` for non-square block sizes). This is something that came in after the first NEON implementations, and I did not know this when I said it was supposed to work in my comment on #453. I checked back with colleagues and we can define that `TARGET_SIMD_X86` is a requirement for `TARGET_SIMD_ARM`.
> 
> So, even with your changes applied, the code does not compile when `TARGET_SIMD_ARM` is enabled, but `TARGET_SIMD_X86` is not. Maybe we should explicitly catch that in the configure step.

Thanks for pointing out the problems with the `xCalcHAD` functions, I now realise that CMake cache variables meant that my existing build directories wasn't actually picking up my attempts to recompile with the SIMDe kernels disabled.

I think it is still possible to avoid the dependency between the two options: we can simply disable the RdCost optimisations if `TARGET_SIMD_X86` is not enabled. This is similar to how we already require AArch64 for these kernels to be used. I've put up a patch to that effect (and swapped the existing `__ARM_ARCH >= 8` check for `REAL_TARGET_AARCH64` now that we have it). Let me know what you think about this approach!

I also have a patch to remove the AArch64-only requirement for the RdCost kernels such that it also compiles on 32-bit Arm platforms, but I will put that up as a separate PR once this is merged so it doesn't conflict.

K-os (approved):
Ok, looks good.

source/Lib/CommonLib/arm/RdCostARM.h
```diff
@@ -1015,16 +1017,16 @@ void RdCost::_initRdCostARM()
   m_afpDistortFunc[0][DF_HAD128_fast]  = RdCost::xGetHADs_ARMSIMD<vext, true>;
 }
 
-#else    // !__ARM_ARCH >= 8
+#else  // !defined( TARGET_SIMD_X86 ) || !REAL_TARGET_AARCH64
```
adamjw24, line 1020: Hmm, the SIMD function for the SADx5 could still be used, even if `TARGET_SIMD_X86` is not defined.

PR #459, opened 2024-11-05
Add an initial unit test framework and add to CTest

### LINKED ISSUE #451: Unit testing of SIMD kernels
When writing AArch64 Neon and SVE implementations for some of the MCTF kernels I noticed that it was quite easy to find cases (e.g. tail loops) which do not appear to be covered by the existing encoding tests run by CTest.

It doesn't appear that there are any existing unit tests for common kernels, so I was wondering if there was appetite to add such test cases to the library? I'd be happy to write test cases for some of the kernels that I have been looking at if you could let me know a preference in terms of test framework library or how best to integrate it with the existing test code?

Thanks!

georges-arm:
Add some initial files and CMake framework for unit testing of individual SIMD kernels with specialized implementations.

For now this is just a simple main program that runs a range of random inputs for two sets of interesting TCoeffOps kernels. By having two instantiations of the TCoeffOps class and only calling initTCoeffOps{ARM,X86} on one of the two class instances, we can effectively compare the reference C/C++ implementation against any optimized kernel if present.

One suggestion for future work may be to test all supported SIMD kernel variants at once rather than just the single one currently selected by run-time CPU feature detection. This would allow us to for instance test both the SIMDe and Arm versions of a kernel without needing to maintain multiple build directories.

Fixes #451

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -375,12 +375,14 @@ if( VVENC_ENABLE_INSTALL )
   endif()
 endif()
 
+set( VVENC_LIB_DIR "${CMAKE_CURRENT_SOURCE_DIR}/source/Lib" )
 
 add_subdirectory( "source/Lib/vvenc" )
 add_subdirectory( "source/App/vvencapp" )
 add_subdirectory( "source/App/vvencFFapp" )
 add_subdirectory( "test/vvenclibtest" )
 add_subdirectory( "test/vvencinterfacetest" )
+add_subdirectory( "test/vvenc_unit_test" )
 
 include( cmake/modules/vvencTests.cmake )
 
--- a/cmake/modules/vvencTests.cmake
+++ b/cmake/modules/vvencTests.cmake
@@ -19,6 +19,8 @@ add_test( NAME Test_vvenclibtest-sdk_default             COMMAND vvenclibtest 4 
 add_test( NAME Test_vvenclibtest-sdk_stringapi_interface COMMAND vvenclibtest 5 )
 add_test( NAME Test_vvenclibtest-timestamps              COMMAND vvenclibtest 6 )
 
+add_test( NAME Test_vvenc_unit_test COMMAND vvenc_unit_test )
+
 set( CLEANUP_TEST_FILES "" )
 
 function( add_vvenc_test NAME TIMEOUT OUT_OUTPUT REQUIRES )
--- a/source/Lib/CommonLib/CommonDef.h
+++ b/source/Lib/CommonLib/CommonDef.h
@@ -543,8 +543,8 @@ struct ClpRng
 struct ClpRngs : public ClpRng
 {
   // bitdepth has to be the same for all channels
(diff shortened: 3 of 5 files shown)


georges-arm:
Keep in mind this is only really intended as a first attempt or proof of concept, but I'd be happy to iterate on this towards something more suitable or add additional functionality if you think it would be useful to have in the repo.

In particular I've tried to avoid adding a load of command line options to control e.g. what tests to run, or what architecture 
features to enable, or how many test cases to run, or to select a random seed, but I imagine that eventually it would be useful to have some of these as a later PR.

Thanks!

adamjw24:
Any idea how to fix the windows build?

Otherwise it looks good.

georges-arm:
> Any idea how to fix the windows build?

So seems like fixing up the unused variables got the MSVC builds working again, but the MinGW build is still broken with an error on the declaration of `g_msgFnc`:

```In file included from D:/a/vvenc/vvenc/source/Lib/CommonLib/TrQuant_EMT.h:48,
                 from D:\a\vvenc\vvenc\test\vvenc_unit_test\vvenc_unit_test.cpp:54:
D:/a/vvenc/vvenc/source/Lib/CommonLib/CommonDef.h:571:62: error: ignoring attributes on template argument 'void(void*, int, const char*, va_list)' {aka 'void(void*, int, const char*, char*)'} [-Werror=ignored-attributes]
  571 | extern std::function<void( void*, int, const char*, va_list )> g_msgFnc;
      |                                                              ^
```

I'd assume this is something to do with the `va_list` parameter being special, but I'm not sure why this only seems to get picked up in the new unit test build rather than anywhere else in the library. Will continue investigating.

adamjw24:
@lehmann-c @K-os any idea? I feel like we've seen that one before...

K-os:
Yes, this is how to "fix" it:
```patch
diff --git a/test/vvenc_unit_test/CMakeLists.txt b/test/vvenc_unit_test/CMakeLists.txt
index f91d46cbe..37a370895 100644
--- a/test/vvenc_unit_test/CMakeLists.txt
+++ b/test/vvenc_unit_test/CMakeLists.txt
@@ -13,7 +13,7 @@ if( VVENC_LIBRARY_ONLY )
 endif()
 
 target_compile_options( ${EXE_NAME} PRIVATE $<$<OR:$<CXX_COMPILER_ID:Clang>,$<CXX_COMPILER_ID:AppleClang>>:-Wall>
-                                            $<$<CXX_COMPILER_ID:GNU>:-Wall -fdiagnostics-show-option>
+                                            $<$<CXX_COMPILER_ID:GNU>:-Wall -fdiagnostics-show-option -Wno-ignored-attributes>
                                             $<$<CXX_COMPILER_ID:MSVC>:/W4 /WX /wd4244 /wd4251 /wd4996>)
 
 target_include_directories( ${EXE_NAME} PRIVATE ${VVENC_LIB_DIR} )

```

PR #460, opened 2024-11-07
add skip bt and skip qt condition

wenglihong:


adamjw24:
What is the purpose of this?

I mean, there is no results, you disable a lot of carefully fine tuned logic, disable crucial VVC tools (tripple split). Some comment would be appreciated. I'd probably close it later today.

PR #468, opened 2024-11-18
Fix SIMD unit tests for transforms

adamjw24:
* ensure only actually occuring cases are tested
* ensure only bitdepths are used that actually occur in the codec
  * signed 8 bit for transformation
  * signed 16 bit for coefficients
  * signed 11 bit for residual

### DIFF
--- a/test/vvenc_unit_test/vvenc_unit_test.cpp
+++ b/test/vvenc_unit_test/vvenc_unit_test.cpp
@@ -50,6 +50,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <stddef.h>
 #include <stdio.h>
 #include <vector>
+#include <time.h>
 
 #include "CommonLib/TrQuant_EMT.h"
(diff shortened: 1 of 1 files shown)


test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -241,11 +282,14 @@ static bool test_TCoeffOps()
 
 int main()
 {
+  unsigned seed = rand();
+  srand( seed );
```
K-os, line 286: I think this makes no sense. It gets a random number seeded with 1, and then seeds the random number generator with that "random" value, so it will always return the same sequence of random numbers. If you want to have a different sequence each time the tests run, you usually seed it with the current timestamp.

adamjw24: The initial seed is always 1? Didnt know that. Changed to a `time` call. Thanks

PR #474, opened 2024-11-25
Add AArch64 SVE implementation for TCoeffOps fastFwdCore_2D 

georges-arm:
A bit more CMake framework code to support building SVE code, followed by an initial bit of SVE code that makes use of this infrastructure. This commit does not attempt to set SVE as a default-enabled target for now.
 
### Wire up CMake to enable compiling files with AArch64 SVE/SVE2

This commit enables compilation for new `sve/` and `sve2/` subdirectories under `CommonLib/arm/`, assuming that the appropriate `-march=...` flags are available.

Introduce a new `set_if_compiler_supports_arm_extensions` function to set up variables based on whether the compiler supports SVE and SVE2, then use these along with the existing requirements (SVE2 requires SVE to be enabled, SVE requires Neon to be enabled) to determine what features ultimately end up being enabled.

The implementation of the extension checking helper function needs to check (a) whether the flag is supported, and (b) whether code will successfully compile when using that flag. The latter is needed since there are some old versions of LLVM that are missing the `arm_neon_sve_bridge.h` header, and LLVM currently fails to compile SVE code when targeting Windows.

### Add AArch64 SVE implementation for `TCoeffOps` `fastFwdCore_2D`

The SVE 16-bit dot-product instructions allow us to accumulate twice as much data per instruction compared to Neon multiply-add instructions, giving a good speedup for the `fastFwdCore_2D` kernels.

Compared to Neon with a fixed vector length of 128 bits, SVE allows different micro-architectures to expose a number of different vector lengths: 128, 256, 512, 1024, or 2048 bits. To take advantage of this we can rewrite the innermost loop of `fastFwdCore_2D` to be expressed in terms of the number of vectors to process rather than the number of elements, and then pick the number of iterations at setup-time by inspecting the vector length. This allows us to largely avoid needing an entire set of kernels for each possible vector length.

One caveat to the notion of having completely vector-length agnostic kernels is that when the vector-length is known to be exactly 128-bits (the same as Neon) we can make use of some Neon instructions to speed up processing the data after the accumulation. This is possible since Neon and SVE registers share the low 128-bits of each vector register.

For this commit we have not attempted to add kernels that process less than a full vector's worth of data per inner loop iteration, which would enable using these kernels on machines with very wide vectors (512, 1024, or 2048 bits). This is technically straightforward since SVE supports partial vectors via predication, however there are no known long-vector micro-architectures available at present to justify maintaining such code.

Running a video encoding job on SVE-capable machines using the `--preset=fast` setting shows the following improvements in reported FPS:

    Neoverse V1 (VL=256 bits): ~1.3%
    Neoverse V2 (VL=128 bits): ~2.6%

### DIFF
--- a/cmake/modules/vvencCompilerSupport.cmake
+++ b/cmake/modules/vvencCompilerSupport.cmake
@@ -1,5 +1,6 @@
 include( CheckCCompilerFlag )
 include( CheckCSourceCompiles )
+include( CheckCXXSourceCompiles )
 
 function( set_if_compiler_supports_flag output_var flag )
   string( REGEX REPLACE "[-.]" "_" SUPPORTED_flag_var "SUPPORTED${flag}" )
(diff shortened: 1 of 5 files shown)


cmake/modules/vvencCompilerSupport.cmake
```diff
@@ -84,6 +85,73 @@ function( _emscripten_enable_wasm_simd128 )
...
+  if( NOT(( ${CMAKE_SYSTEM_PROCESSOR} MATCHES "arm64" ) OR
+          ( ${CMAKE_SYSTEM_PROCESSOR} MATCHES "aarch64" )))
+    return()
+  endif()
+  if( UNIX OR MINGW )
+    # SVE is an optional feature from Armv8.2-A.
+    set( ${flag_sve} "-march=armv8.2-a+sve" )
+    set( ${flag_sve} "-march=armv8.2-a+sve" PARENT_SCOPE )
+    _set_if_compiler_supports_sve_flag( ${output_var_sve} "${${flag_sve}}" )
+    set( ${output_var_sve} "${${output_var_sve}}" PARENT_SCOPE )
```
K-os, line 146: This looks overly complicated. Do I understand correctly? You create a *local* variable with the name provided as a function parameter to `set_if_compiler_supports_arm_extensions()` and store the variable *name* in `flag_sve`. You then pass the name of that variable to `_set_if_compiler_supports_sve_flag()` to be overwritten? I think, if you introduced a local variable with a different name, all these complex indirections can be avoided.

Wouldn't something like this also work? Untested:
```cmake
    set( _flag_sve "-march=armv8.2-a+sve" )
    _set_if_compiler_supports_sve_flag( _sve_supported "${_flag_sve}" )
    if( _sve_supported )
      set( ${flag_sve} "${_flag_sve}" PARENT_SCOPE )
      set( ${output_var_sve} "${_sve_supported}" PARENT_SCOPE )
    endif()
```

additionally, if I am not mistaken, the variable in `flag_sve` will always contain `-march=armv8.2-a+sve` regardless of whether it is supported or not.

My function `set_if_compiler_supports_flag()` is intended to work like this:
You pass an output variable and a flag, and if the compiler does support the flag, the output variable will contain the flag, so you can just use the variable in the compile options. I would prefer if your similarly named function behaved somehow similar.
It's fine for me if you like to add a separate boolean output variable to explicitly convey whether SVE is supported, but at least in respect to the flag variable it should behave similar.

source/Lib/vvenc/CMakeLists.txt
```diff
@@ -133,7 +135,35 @@ endif()
 if( VVENC_ENABLE_ARM_SIMD )
   # Set needed compile definitions.
   set_property( SOURCE ${ARM_NEON_SRC_FILES} APPEND PROPERTY COMPILE_DEFINITIONS USE_NEON )
-  add_library( ${LIB_NAME}_arm_simd OBJECT ${ARM_NEON_SRC_FILES} )
+
+  if(( ${CMAKE_SYSTEM_PROCESSOR} MATCHES "arm64" ) OR ( ${CMAKE_SYSTEM_PROCESSOR} MATCHES "aarch64" ))
+    # Neon is mandatory in AArch64, so no additional compile flags needed here.
+  else()
+    set_property( SOURCE ${ARM_NEON_SRC_FILES} APPEND PROPERTY COMPILE_FLAGS "-mfpu=neon" )
```
K-os, line 142: This will not work with Visual Studio. Please use `set_if_compiler_supports_flag()` first.

K-os (changes requested):

source/Lib/vvenc/CMakeLists.txt
georges-arm: Done!

cmake/modules/vvencCompilerSupport.cmake
georges-arm: Thanks for the suggestions, that does simplify things quite a bit.

PR #475, opened 2024-11-25
AArch32 and AArch64 minus SIMDe fixes

georges-arm:
It seems like #473 broke the VVenC build when building both for AArch32 and also for AArch64 if SIMDe-based kernels are not enabled. This is easily fixed through a few small tweaks:

* Use the existing helper functions in `sum_neon.h` and avoiding `*_high_*` and other intrinsics that are not available for 32-bit Arm platforms.

* Guard the SIMDe-enabled kernels with `defined( TARGET_SIMD_X86 )` checks to avoid including these kernels if they are not available.

* Additionally fix a typo in `InterPredARM.h` to refer to void `InterPredInterpolation` instead of `TCoeffOps`.

### DIFF
--- a/source/Lib/CommonLib/arm/InterPredARM.h
+++ b/source/Lib/CommonLib/arm/InterPredARM.h
@@ -274,10 +274,10 @@ void InterPredInterpolation::_initInterPredictionARM()
   xFpBiDirOptFlow     = BiOptFlowCoreARMSIMD<vext>;
 }
 
-#else 
+#else
 
 template<ARM_VEXT vext>
-void TCoeffOps::_initInterPredictionARM()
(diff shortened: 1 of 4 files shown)


georges-arm:
Hi @adamjw24 I'm not sure what the current CI/pre-commit setup looks like, do you think it would be possible to add some additional checks to ensure that compilation on 32-bit and 64-bit Arm + the various CMake SIMD configure option combinations continues to work?

adamjw24:
While its a fair request, I think this is a very artificial use-case. In addition to the GitHub CI, we have an internal runner with around 30 builds, 10 of them also running the tests. I think stuff like here will happen occasionaly and can than be fixed. We'll put more considration into the use case the next time tho.

PR #476, opened 2024-11-26
Use ARM NEON code even if SIMDe is disabled

adamjw24:
@K-os could you confirm this compiles and runs on ARM with and without SIMDe (i.e. TARGET_SIMD_X86=0)?


Proudsalsa:
I tested it on a M2 machine and both configurations pass make test.

PR #478, opened 2024-12-03
Add unit test for checking jumps backwards in DTS

carlosfg-flu:
This simple test checks for any jump backward in the DTS that the encoder outputs.

Right now the test is failing:
```sh
$ make debug && bin/debug-static/vvencdtstest
...
  5
 -4	[jump backwards]
 -3
 -2
 -1
  0
  1
  2
  3
 14
  5	[jump backwards]
  6
  7
 18
 19
 20
 21
 22
 23
 24
 $ echo $?
 1
 ```


adamjw24:
It would make sense to add this test to `make test`. We'll have a look at the behavior itself. It would also be nice if the DTS testcase contained more than just one single testcase. Maybe if thats just one testcase it would be easier to move it to library test?

PR #511, opened 2025-02-23
Add AArch64 CPU feature detection for OpenBSD and FreeBSD

brad0:


source/Lib/CommonLib/arm/CommonDefARM.cpp
```diff
@@ -45,8 +45,8 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "CommonDefARM.h"
 
-#if defined( __linux__ )
-#include <sys/auxv.h>  // getauxval
+#if defined( __linux__ ) || HAVE_ELF_AUX_INFO
+#include <sys/auxv.h>  // getauxval / elf_aux_info
 #endif
```
K-os, line 50: I know it was like that before, but  `getauxval()` / `elf_aux_info()` are called only, when `TARGET_SIMD_ARM_SVE` or `#if TARGET_SIMD_ARM_SVE2` is set. Would it make sense to only include `sys/auxv.h` when these macros are set?

In that case, I would prefer to make the corresponding call to `check_symbol_exists()` also dependent on SVE/SVE2.

K-os (changes requested):
Thanks for the update.
Can you please fix the indentation in the CMakeLists.txt? Then it looks good to me.

brad0:
> Thanks for the update. Can you please fix the indentation in the CMakeLists.txt? Then it looks good to me.

Fixed.

K-os (approved):
Thanks. Looks good.

PR #513, opened 2025-03-10
Support passing through user data per frame

### LINKED ISSUE #480: Allow user data pass through
It would be useful to link some user data to an incoming frame and be able to recover it at the output. This will make integration with frameworks like FFmpeg [1] and GStreamer [2] more seamless. Other encoder libraries do support this [3].

[1]: FFmpeg has this [`opaque`](https://github.com/FFmpeg/FFmpeg/blob/6057cd8aeee5c2ed8397fcbf0ee74de26701b5aa/libavcodec/packet.h#L561-L564) feature
[2]: GStreamer has [`GstMeta`](https://gstreamer.freedesktop.org/documentation/gstreamer/gstmeta.html?gi-language=c), used by [`GstVideoEncoder`](https://gstreamer.freedesktop.org/documentation/video/gstvideoencoder.html) base class.
[3]: Field [`userData`](https://github.com/videolan/x265/blob/master/source/x265.h#L404-L407) in x265 encoder

cadubentzen:
This is guarded under `#ifdef VVENC_USE_UNSTABLE_API` and disabled by default due to requiring breaking changes to the ABI. It can only be unguarded when the major version is bumped to v2.
    
Closes #480

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -179,6 +179,8 @@ set( VVENC_INSTALL_FULLFEATURE_APP          OFF CACHE BOOL   "Install the full-f
 
 set( VVENC_ENABLE_WERROR                    ON  CACHE BOOL   "Treat warnings as errors (-Werror or /WX)" )
 
+set( VVENC_ENABLE_UNSTABLE_API              OFF CACHE BOOL   "Enable unstable API" )
+
 if( CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR
     CMAKE_CXX_COMPILER_ID STREQUAL "Clang" )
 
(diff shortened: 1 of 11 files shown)


cadubentzen:
cc @adamjw24

adamjw24:
Hi @cadubentzen - we are aware the solution to the problem is as simple as that, but it changes the interface a lot.

This is a very non-trivial change. It will not be ABI-compatible to 1.13.0. Such a change might actually only be possible for 2.0.0 which we are currently not planing.

Any idea how this could be done without breaking ABI compatibility?

cadubentzen:
Thanks for the reply!

Without breaking ABI, this is tricky indeed. That's one of the reasons libraries add padding to struct members to allow for changes like this. For v2, it would be wise add such padding as well to be more future-proof.

One way to avoid breaking ABI compatibility now while still allow for ABI breaking changes is to put those changes under a `#ifdef VVENC_USE_UNSTABLE_API` guard for those new struct members, which is disabled by default and unguarded only when the major version is bumped. I can adapt my patch for that if you think this is sensible.

adamjw24:
Yes, thats a good idea.

Please:
* rename `opaque` to `userData`
* wrap the changes in the define `VVENC_USE_UNSTABLE_API`
* add the switch to CMake and Makefile to be able to control during build (default off)

cadubentzen:
@adamjw24 done

cadubentzen:
No problem 😊 Thanks a lot for looking into this carefully. That's a very good point! Will do.

adamjw24:
> Done

Thanks!

Issue #519, opened 2025-03-14
old nlohmann json in thirdparty causing compilation warning/error with clang


1480c1:
```c++
FAILED: source/Lib/vvenc/CMakeFiles/vvenc.dir/vvencCfg.cpp.o 
ccache D:\clang\msys64\clang64\bin\ccache.exe  clang++ -DHAVE_PTHREADS -DTARGET_SIMD_X86=1 -DVVENC_SOURCE -D_WIN32_WINNT=0x0600 -ID:/clang/build/vvenc-git/source/Lib/vvenc/../../../include -ID:/clang/build/vvenc-git/build-64bit -ID:/clang/build/vvenc-git/source/Lib/vvenc/. -ID:/clang/build/vvenc-git/source/Lib/vvenc/.. -ID:/clang/build/vvenc-git/source/Lib/vvenc/../DecoderLib -ID:/clang/build/vvenc-git/source/Lib/vvenc/../EncoderLib -ID:/clang/build/vvenc-git/source/Lib/vvenc/../CommonLib -ID:/clang/build/vvenc-git/source/Lib/vvenc/../CommonLib/x86 -ID:/clang/build/vvenc-git/source/Lib/vvenc/../CommonLib/arm -ID:/clang/build/vvenc-git/source/Lib/vvenc/../apputils -ID:/clang/build/vvenc-git/source/Lib/vvenc/../../../thirdparty/nlohmann_json/single_include -isystem D:/clang/build/vvenc-git/source/Lib/vvenc/../../../thirdparty -D_FORTIFY_SOURCE=2 -fstack-protector-strong -mtune=generic -O2 -pipe -D__USE_MINGW_ANSI_STDIO=1 -DVVENC_ENABLE_THIRDPARTY_JSON -DVVENC_ENABLE_THIRDPARTY_JSON -O3 -DNDEBUG -std=gnu++14 -fvisibility=hidden -fvisibility-inlines-hidden -Werror -Wall -Wno-deprecated-register -Wno-unused-const-variable -Wno-unknown-attributes -MD -MT source/Lib/vvenc/CMakeFiles/vvenc.dir/vvencCfg.cpp.o -MF source\Lib\vvenc\CMakeFiles\vvenc.dir\vvencCfg.cpp.o.d -o source/Lib/vvenc/CMakeFiles/vvenc.dir/vvencCfg.cpp.o -c D:/clang/build/vvenc-git/source/Lib/vvenc/vvencCfg.cpp
[...]

PR #520, opened 2025-03-19
json: add support for using an external nlohmann_json

1480c1:
Closes https://github.com/fraunhoferhhi/vvenc/issues/519

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,7 +173,9 @@ set( VVENC_ENABLE_BUILD_TYPE_POSTFIX        OFF CACHE BOOL   "Enable or disable 
 
 set( VVENC_ENABLE_LINK_TIME_OPT             ON  CACHE BOOL   "Enable link time optimization for release and profile builds" )
 
-set( VVENC_ENABLE_THIRDPARTY_JSON           ON  CACHE BOOL   "Enable use of thirdparty json library" )
+set( VVENC_ENABLE_THIRDPARTY_JSON           ON  CACHE STRING "Enable use of thirdparty json library, pass in SYSTEM to rely on an external nlohmann_json" )
+
+set_property( CACHE VVENC_ENABLE_THIRDPARTY_JSON PROPERTY STRINGS OFF SYSTEM ON )
 
 set( VVENC_INSTALL_FULLFEATURE_APP          OFF CACHE BOOL   "Install the full-feature app: vvencFFapp" )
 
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -173,7 +173,9 @@ set( VVENC_ENABLE_BUILD_TYPE_POSTFIX        OFF CACHE BOOL   "Enable or disable
 
 set( VVENC_ENABLE_LINK_TIME_OPT             ON  CACHE BOOL   "Enable link time optimization for release and profile builds" )
 
-set( VVENC_ENABLE_THIRDPARTY_JSON           ON  CACHE BOOL   "Enable use of thirdparty json library" )
+set( VVENC_ENABLE_THIRDPARTY_JSON           ON  CACHE STRING "Enable use of thirdparty json library, pass in SYSTEM to rely on an external nlohmann_json" )
```
adamjw24, line 176: It seems changing the variable type to `STRING` confuses CMake a lot. In string logic "OFF" is also true, so I don't think this can work as you wanted. Maybe a new variable would be better after all.

CMakeLists.txt
```diff
@@ -368,12 +370,6 @@ if( CCACHE_FOUND )
   set_property( GLOBAL PROPERTY RULE_LAUNCH_LINK ccache )
 endif()
 
-
-if( VVENC_ENABLE_THIRDPARTY_JSON )
-  set( CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DVVENC_ENABLE_THIRDPARTY_JSON" )
-  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DVVENC_ENABLE_THIRDPARTY_JSON" )
-endif()
```
1480c1, line 375: I'm starting to see why this was duplicated here.

adamjw24:
Since you're on it.

Could you also change in the Makefile:

```Makefile
ifneq ($(disable-json),)
CONFIG_OPTIONS += -DVVENC_ENABLE_THIRDPARTY_JSON=OFF
endif
```

to


```Makefile
ifneq ($(enable-json),)
CONFIG_OPTIONS += -DVVENC_ENABLE_THIRDPARTY_JSON=${enable-json}
endif
```

so that we can set this option using `make`?

CMakeLists.txt
adamjw24: Good catch!

adamjw24:
I like the solution, thanks!

PR #522, opened 2025-03-21
nlohmann_json: Fix linking for downstream projects

1480c1:
FFmpeg and libheif were failing to link with the new changes for linking an external nlohmann_json for separate reasons.

For FFmpeg, the reason was that the pkg-config file contained invalid options such as

```bash
Libs.private: ... -lkernel32 interface_libs-NOTFOUND interface_libs-NOTFOUND -lpthread
```

Which, of course, FFmpeg can't find `interface_libs-NOTFOUND`.
This occurred due to a few coincidences:

1. The `interface_libs` was populated with `$<LINK_ONLY:nlohmann_json::nlohmann_json>;$<LINK_ONLY:Threads::Threads>`, etc..
2. `resolve_target_interface_libs()` would see that `$<LINK_ONLY:nlohmann_json::nlohmann_json>` was a generator expression, extracting the inner library and resolving it,
3. Since `nlohmann_json::nlohmann_json` doesn't actually have any backing library, it would return `interface_libs-NOTFOUND`
4. Depending on the compiler and system, `Threads::Threads` could return `interface_libs-NOTFOUND`, or could return `-lpthread`. In the case of MSYS2, mingw64 gcc returned `interface_libs-NOTFOUND`, clang64 clang returned `-lpthread`.
5. The resulting `VVENC_PKG_INTERFACE_LIBS` contained `interface_libs-NOTFOUND;interface_libs-NOTFOUND;-lpthread` for clang, and `interface_libs-NOTFOUND;interface_libs-NOTFOUND;interface_libs-NOTFOUND` for gcc. 
6. The `if( VVENC_PKG_INTERFACE_LIBS )` would fail for gcc's output as it ends in `-NOTFOUND`, but would append for clang.

This resulted in `libvvenc.pc` compiled by clang64's clang would fail to link.

The fix is to just check if `interface_libs` contains a valid-ish library (something that at least passes cmake's `if` string checks).

For libheif, the resulting `vvencTargets-static.cmake` contained the line:

```cmake
INTERFACE_LINK_LIBRARIES "\$<LINK_ONLY:nlohmann_json::nlohmann_json>;\$<LINK_ONLY:Threads::Threads>"
```

This caused issues, as any downstream user like libheif would need to provide `nlohmann_json::nlohmann_json` themselves (either by `find_package()` or adding it as an imported library manually).

I assume that normally the cmake file would contain something like `find_package(nlohmann_json)`.

However, in the case of vvenc, I believe it's unnecessary as the resulting library doesn't actually need its nlohmann_json dependency exposed since the public headers don't include it and there's nothing to link to. Thus, we can work around it by imitating what `target_link_libraries()` did by just adding the cflags where it's needed.


cmake/modules/vvencInstall.cmake
```diff
@@ -111,7 +111,9 @@ function( resolve_target_interface_libs TGT OUT_VAR )
       resolve_target_interface_libs( ${lib} lib )
     endif()
 
-    list( APPEND ret ${lib} )
+    if( lib )
+      list( APPEND ret ${lib} )
+    endif()
```
adamjw24, line 116: Should this change not be enough to fix the problem?

adamjw24 (commented):
Could you elaborate a bit? We tried to reproduce the problem with Ubuntu without luck, which I guess is why we missed it in the first place.

Does the problem occur when you use either SYSTEM or ON, or only when you use SYSTEM for the JSON?

Best, Adam

1480c1:
> Could you elaborate a bit? We tried to reproduce the problem with Ubuntu without luck, which I guess is why we missed it in the first place.
> 
> Does the problem occur when you use either SYSTEM or ON, or only when you use SYSTEM for the JSON?
> 
> Best, Adam


This occurs mainly on Windows.
The pkg-config file issue happens only with clang

|              | pkg-config | cmake |
|--------------|------------|-------|
| gcc SYSTEM   |            | x     |
| gcc  ON      |            | x     |
| clang SYSTEM | x          | x     |
| clang ON     | x          | x     |

So the `if ( lib )` change only fixes the pkg-config issue on clang.

The "json: don't directly link to target" commit helps with gcc and clang. Its effect is that inside the `vvencTargets-static.cmake` file that is generated by cmake, currently without this PR it shows


```cmake
set_target_properties(vvenc::vvenc PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "\$<\$<STREQUAL:\$<TARGET_PROPERTY:vvenc::vvenc,TYPE>,SHARED_LIBRARY>:VVENC_DYN_LINK>"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "\$<LINK_ONLY:nlohmann_json::nlohmann_json>;\$<LINK_ONLY:Threads::Threads>"
  INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "include"
)
```

And the `\$<LINK_ONLY:nlohmann_json::nlohmann_json>;` line is the main issue with linking to vvenc from libheif, as it will cause cmake to try to find `nlohmann_json::nlohmann_json`, but since libheif doesn't do any kind of `find_package(nlohmann_json)`, it won't resolve. The commit addresses this by doing the same effect as what we had with `nlohmann_json`, but without adding it to `INTERFACE_LINK_LIBRARIES`.

cmake/modules/vvencInstall.cmake
1480c1: Answered in a separate comment

adamjw24:
I dont understand - are you including vvenc as a subproject to libheif? Why not just build it properly and link it like you would usually do when using external libraries?

1480c1:
> I dont understand - are you including vvenc as a subproject to libheif? 

No, libheif calls `find_package(vvenc)`
https://github.com/strukturag/libheif/blob/master/CMakeLists.txt#L177


>Why not just build it properly and link it like you would usually do when using external libraries?

We do, and when libheif encounters the `find_package(vvenc)` call, it errors out

adamjw24:
Where do you find the `Findvvenc.cmake` and `Findvvdec.cmake` files that would describe that call? I see libheif has specific definitions for other codecs, but not ours. We do not install any cmake descriptors otehr than the pkg-config.

1480c1:
> Where do you find the `Findvvenc.cmake` and `Findvvdec.cmake` files that would describe that call? I see libheif has specific definitions for other codecs, but not ours. We do not install any cmake descriptors otehr than the pkg-config.

The `vvencConfig.cmake` file and other `.cmake` installed in `cmake/modules/vvencInstall.cmake` provide the files necessary for cmake to find them when `find_package` is called.

adamjw24:
Indeed, thank you.

adamjw24:
Thanks for resolving the issue!

PR #524, opened 2025-03-28
SEI: implement sdi and aci writers

diegonieto:
In order to enable alpha channel signalling, this PR implements the SEI writers, which is a part of needs for that aim:
* ITU 274 8.19: Scalability Dimension Info
* ITU 274 8.23: Alpha Channel


adamjw24:
There is no functionality to this PR. The additional code could never be used cosindering the current VVenC interface and the fact that VVenC does not even support multi-layer.

What is the actual added value of this PR for the end users?

adamjw24:
I see, thank you for the explanation. While I am glad to see that this MR comes from an actual working implementation, I think I would prefer not to take it in as is. We try not to have stuff that is not directly used. Also glad to see you guys keep the ML encoder up to date.

adamjw24:
Btw, I did a patch for the NHK implementation once. Some of the code is now part of VVenC, especially the 1xN and Mx1 interpolation filters, but there are a few options to remove some obsolete multiple checks in the inter prediction as well. Please find attached, maybe its useful for you.

[improve_rpr_for_multilayer.patch](https://github.com/user-attachments/files/19545495/improve_rpr_for_multilayer.patch)

PR #526, opened 2025-04-11
Fix 32-bit Arm build when VVENC_ENABLE_ARM_SIMD is disabled

georges-arm:
It is possible to enable the `VVENC_ENABLE_ARM_SIMD` and `VVENC_ENABLE_X86_SIMD` options independent of each other in the `CMakeLists.txt` configuration, however if only the Arm Neon kernels are disabled the resulting code causes bus errors when built and run under 32-bit Arm.

This is due to a pair of issues:

* The existing `_vv_` unaligned macros are guarded by `TARGET_SIMD_ARM` rather than `REAL_TARGET_ARM`. This means that if the Neon kernels are disabled but the x86 kernels remain enabled, the macros are then incorrectly disabled.

* There is a missing macro covering use of the `_mm_lddqu_si128` intrinsic, which causes tests to crash due to alignment issues. With the `_vv_` macro guard fixed we can adjust this to use `_mm_loadu_si128` on 32-bit Arm targets instead.

### DIFF
--- a/source/Lib/CommonLib/x86/BufferX86.h
+++ b/source/Lib/CommonLib/x86/BufferX86.h
@@ -1888,9 +1888,9 @@ uint64_t AvgHighPass_SIMD( const int width, const int height, const Pel* pSrc, c
       for (x = 1; x < width-1-6; x += 6)
       {
         sum=0;
-        lineM1 = _mm_lddqu_si128 ((__m128i*) &pSrc[ (y -1)  *iSrcStride + x-1]);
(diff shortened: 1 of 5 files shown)


adamjw24:
I would propose to just always use `_mm_loadu_si128`. I don't think there is much reason to use `_mm_lddqu_si128`, and it seems less portable with our methods. This way we could avoid adding one more macro and level of abstraction.

Could you adapt appropriately?

PR #527, opened 2025-04-11
Arm: Enable inter-prediction optimizations with SVE/SVE2 enabled

georges-arm:
The existing code only enables the inter-prediction optimized kernel when the extension flags report exactly Neon. This means that if SVE or SVE2 are returned instead then this kernel will be unintentionally disabled.

Adjust this condition to check if _at least_ Neon is enabled in order to match the other `init*` functions in this file.


adamjw24:
Thanks!

PR #465, opened 2024-11-11
Add AArch64 Neon implementation of motionErrorLumaInt8

georges-arm:
Add a new Neon implementation for the MCTF `motionErrorLumaInt8` kernel, performing a sum of squared difference calculation with early-exit.

Running a video encoding job on a Neoverse V2 machine using the `--preset=fast` setting shows a ~0.5% improvement in reported FPS.

PR #528, opened 2025-04-11
Add unit tests for MCTF motionErrorLumaInt8

georges-arm:
In #465 we added a Neon implementation of motionErrorLumaInt8, however this was not covered by any of the existing unit tests. To try and ensure that the behaviour of the x86 and Arm kernels continues to match, add a unit test covering these kernels.

To enable the tests to call the optimized kernels in a way matching how we previously did this for `TCoeffOps`, mark the MCTF member kernels as public members of the class and pass a new `enableOpt` member to allow us to compare the optimized and fallback scalar implementations.

Also add `-Wno-sign-compare` when compiling unit tests with GCC to avoid warnings from `Common.h` and `Slice.h`, since this flag is already present when building the library.

### DIFF
--- a/source/Lib/CommonLib/MCTF.cpp
+++ b/source/Lib/CommonLib/MCTF.cpp
@@ -545,7 +545,7 @@ double calcVarCore( const Pel* org, const ptrdiff_t origStride, const int w, con
   return variance / 256.0;
 }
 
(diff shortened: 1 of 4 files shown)


adamjw24:
Thanks!

PR #534, opened 2025-04-23
Add unit tests for InterPrediction xFpBiDirOptFlow

georges-arm:
We have both x86 and Arm implementations of this kernel, so add some simple unit tests to ensure that the behaviour matches between them.

This requires us to extract the C fallback implementation for `xFpBiDirOptFlow` to its own function and wire it up as the default function pointer. As part of this we also need to move `xRightShiftMSB` to be a non-member function so it is callable from the new fallback implementation.

Similar to other unit tests, we add a new `enableOpt` parameter to the `InterPredInterpolation::init` function to allow us to construct both a reference and optimized object to compare against in the test.

### DIFF
--- a/test/vvenc_unit_test/vvenc_unit_test.cpp
+++ b/test/vvenc_unit_test/vvenc_unit_test.cpp
@@ -52,6 +52,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <stdio.h>
 #include <time.h>
 
+#include "CommonLib/InterPrediction.h"
 #include "CommonLib/MCTF.h"
 #include "CommonLib/TrQuant_EMT.h"
 #include "CommonLib/TypeDef.h"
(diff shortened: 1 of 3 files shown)


test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -360,6 +362,98 @@ static bool test_MCTF()
...
+  opt->xFpBiDirOptFlow( srcY0.data(), srcY1.data(), gradX0.data(), gradX1.data(), gradY0.data(), gradY1.data(), width,
+                        height, dstYopt.data(), dstStride, shift, offset, limit, clpRng, bitDepth );
+  return compare_values_2d( sstm.str(), dstYref.data(), dstYopt.data(), height, dstStride );
+}
+
+static bool check_biDirOptFlow( InterPredInterpolation* ref, InterPredInterpolation* opt, unsigned num_cases, int width,
+                                int height )
+{
+  printf( "Testing InterPred::xFpBiDirOptFlow w=%d h=%d\n", width, height );
+  InputGenerator<TCoeff> g{ 10 };
```
adamjw24, line 409: Should be `InputGenerator<Pel>`

georges-arm: Good point, done!

adamjw24:
Thanks!

PR #537, opened 2025-04-28
Add unit tests for lumaWeightedSSE

athulya-arm:
Add unit tests for lumaWeightedSSE

Add unit tests for lumaWeightedSSE functions. Modify the RdCost::create
function to include a new 'enableOpt' parameter to create a reference
object and an optimized object to test.

### DIFF
--- a/test/vvenc_unit_test/vvenc_unit_test.cpp
+++ b/test/vvenc_unit_test/vvenc_unit_test.cpp
@@ -54,6 +54,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "CommonLib/InterPrediction.h"
 #include "CommonLib/MCTF.h"
+#include "CommonLib/RdCost.h"
 #include "CommonLib/TrQuant_EMT.h"
 #include "CommonLib/TypeDef.h"
 
@@ -66,8 +67,8 @@ static inline bool compare_value( const std::string& context, const T ref, const
 {
   if( opt != ref )
   {
-    printf( "failed: %s\n", context.c_str() );
-    printf( "  mismatch:  ref=%d  opt=%d\n", ref, opt );
+    std::cerr << "failed: " << context << "\n"
+              << "  mismatch:  ref=" << ref << "  opt=" << opt << "\n";
   }
   return opt == ref;
 }
@@ -99,17 +100,25 @@ template<typename T>
 class InputGenerator
 {
 public:
-  explicit InputGenerator( unsigned bits ) : m_bits( bits )
+  explicit InputGenerator( unsigned bits, bool is_signed = true ) : m_bits( bits ), m_signed( is_signed )
   {
   }
 
   T operator()() const
   {
-    return ( rand() & ( ( 1 << m_bits ) - 1 ) ) - ( 1 << m_bits >> 1 );
+    if( !m_signed )
+    {
+      return static_cast<T>( rand() & ( ( 1 << m_bits ) - 1 ) );
+    }
+    else
+    {
+      return ( rand() & ( ( 1 << m_bits ) - 1 ) ) - ( 1 << m_bits >> 1 );
+    }
   }
 
 private:
   unsigned m_bits;
+  bool m_signed;
 };
 
 template<typename T>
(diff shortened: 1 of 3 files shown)


test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -454,6 +462,80 @@ static bool test_InterPred()
...
+
+  std::ostringstream sstm;
+  sstm << "lumaWeightedSSE" << " w=" << width << " h=" << height;
+
+  bool passed = true;
+  for( unsigned i = 0; i < num_cases; i++ )
+  {
+    DimensionGenerator rng;
+    int stride = rng.get( 1, 1024 );
+    int size = std::max( stride, width );
```
adamjw24, line 478: `size` should always be larger of `width`, since it governs the blocks shapes to be tested. A block of size 3xN cannot exist in the codec and thus does not need to be tested.

test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -454,6 +462,80 @@ static bool test_InterPred()
...
+{
+  RdCost ref;
+  RdCost opt;
+  ref.create( /*enableOpt=*/false );
+  opt.create( /*enableOpt=*/true );
+
+  unsigned num_cases = NUM_CASES;
+  bool passed = true;
+  std::array<int, 8> widths = { 1, 2, 4, 8, 16, 32, 64, 128 };
+  std::array<int, 7> heights = { 2, 4, 8, 16, 32, 64, 128 };
```
adamjw24, line 526: Not sure blocks of width 1 are possible, and luma blocks of width or height 2 either. Also, no 4x4, 4x8 or 8x4 luma BI-predicted blocks are possible.

test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -454,6 +462,80 @@ static bool test_InterPred()
...
+    dtParam.org.width = width;
+    dtParam.org.height = height;
+    dtParam.cur.stride = stride;
+    dtParam.org.stride = stride;
+    CPelBuf pelBuf;
+    pelBuf.buf = orgLumaBuf.data();
+    pelBuf.stride = stride;
+    dtParam.orgLuma = &pelBuf;
+    dtParam.bitDepth = 10;
+    dtParam.compID = i % 2 ? COMP_Y : COMP_Cb;
```
adamjw24, line 496: If `num_cases == 1` no chroma test is performed. I think `num_cases` should not be used to control how to perform a test. But than, the tested `lumaWeightedSSE`, might not need be tested for chroma.

test/vvenc_unit_test/vvenc_unit_test.cpp
adamjw24: Forget this comment - I see this function is also used during search and also for funny sizes like 1xN.

test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -105,7 +106,14 @@ class InputGenerator
 
   T operator()() const
   {
-    return ( rand() & ( ( 1 << m_bits ) - 1 ) ) - ( 1 << m_bits >> 1 );
```
adamjw24, line 108: I am not sure this will work here, since `Pel` is used for both signal values (which are unsigned) but also for residual values (which are signed). I think maybe a different approach would be preffered with a flag passed in with the `InputGenerator` to indicate signedness of the generated values.

test/vvenc_unit_test/vvenc_unit_test.cpp
athulya-arm: stride is set as
int stride = rng.get( width, 1024 );

test/vvenc_unit_test/vvenc_unit_test.cpp
athulya-arm: Done.

test/vvenc_unit_test/vvenc_unit_test.cpp
adamjw24: I am sorry, I think I was not too clear. Thanks you for separating num_cases from deciding on what to test. But since the function tested is only meant for luma signals, the following can always be used: `dtPAram.compID = COMP_Y`

adamjw24:
Thanks!

PR #548, opened 2025-05-29
Arm: Add Neon implementation of EqualCoeffComputer and add min/max input to its unit test

micro-arm:
Add min or max input unit test for EqualCoeffComputer
- Add min or max input unit test to EqualCoeffComputer to catch any
mismatch due to bit overflows. Also pull out the height and width into
for loops to test all possible height-width combinations.

Arm: Add Neon implementation of EqualCoeffComputer
- Wire up initAffineGradientSearchARM and add simdEqualCoeffComputer_neon.
- This Neon intrinsics implementation is 3.89x faster (false case) and
3.18x faster (true case) than the SIMDe implementation when benchmarked
on Neoverse N/V-series microarchitectures using LLVM 20.

### DIFF
--- a/test/vvenc_unit_test/vvenc_unit_test.cpp
+++ b/test/vvenc_unit_test/vvenc_unit_test.cpp
@@ -114,6 +114,40 @@ static inline bool compare_values_2d( const std::string& context, const T* ref, 
 }
 
 template<typename T>
+class MinMaxGenerator
+{
+public:
+  explicit MinMaxGenerator( unsigned bits, bool is_signed = true )
+  {
+    if( is_signed )
+    {
+      T half = 1 << ( bits - 1 );
+      m_min = -half;
+      m_max =  half - 1;
+    }
+    else
+    {
+      m_min = 0;
+      m_max = ( 1 << bits ) - 1;
+    }
+  }
+
+  T operator()() const
+  {
+    return ( rand() & 1 ) ? m_max : m_min;
+  }
+
+  std::string input_type() const
+  {
+    return "MinOrMax";
+  }
+
+private:
+  T m_min;
+  T m_max;
+};
+
+template<typename T>
 class InputGenerator
 {
 public:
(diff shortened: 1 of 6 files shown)


test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -690,40 +720,44 @@ static bool check_EqualCoeffComputer( AffineGradientSearch* ref, AffineGradientS
...
     std::ostringstream sstm_test;
-    sstm_test << "AffineGradientSearch::EqualCoeffComputer" << " b6Param=" << std::boolalpha
-              << static_cast<bool>( b6Param );
+    sstm_test << "AffineGradientSearch::EqualCoeffComputer<" << std::boolalpha << static_cast<bool>( b6Param ) << ">"
+              << " Input=" << inp_gen.input_type();
     std::cout << "Testing " << sstm_test.str() << std::endl;
 
-    for( unsigned n = 0; n < num_cases; n++ )
+    // Set height and width to powers of two >= 8.
+    for( int height : { 8, 16, 32, 64, 128 } )
```
adamjw24, line 728: In VVC, affine is only allowed for width > 8 && height > 8, i.e. no size can be smaller than 16. Could you adapt?

test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -113,6 +113,30 @@ static inline bool compare_values_2d( const std::string& context, const T* ref,
   return true;
 }
 
+template<typename T>
+class FixedInputGenerator
+{
+public:
+  explicit FixedInputGenerator( T val ) : m_val( val )
```
adamjw24, line 120: I dont like the idea of a fixed value generator. If you really want to test overflows, I'd say doing patterns of min/max values would be more suitable (as the search often does operation on differences of values). Maybe a min-max generator would be more interesting, that randomly selects either the min or max values.

PR #550, opened 2025-05-30
Add unit tests for applyPlanarCorrection

athulya-arm:
Add unit tests for MCTF::applyPlanarCorrection functions.

### DIFF
--- a/test/vvenc_unit_test/vvenc_unit_test.cpp
+++ b/test/vvenc_unit_test/vvenc_unit_test.cpp
@@ -391,6 +391,55 @@ static bool test_TCoeffOps()
 
 #if ENABLE_SIMD_OPT_MCTF
 template<typename G>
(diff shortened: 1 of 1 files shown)


test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -443,6 +492,11 @@ static bool test_MCTF()
   unsigned num_cases = NUM_CASES;
   bool passed        = true;
 
+  for( int size = 4; size <= 32; size *= 2 )
```
adamjw24, line 495: Could you separate out width and height (and test `{ 4, 8, 16, 24, 32 }`? On picture boundaries we may have strange block sizes.

athulya-arm: It seems like width value 24 does not work for x86 AVX2 because the loop for width is unrolled for 16 when w > 8. Could you please confirm if we are expecting width to be 24?

Issue #562, opened 2025-06-17
addAvg (no stride) SIMD implementations do not handle non-power-of-two sizes


micro-arm:
When running a real-world encoding, we actually found some non-power-of-two sizes that gets passed to addAvg (no stride) function.

On Arm, this causes a CHECK failure: 
```
terminate called after throwing an instance of 'vvenc::Exception'
  what():  ERROR: In function "addAvg_neon" in /home/microb03/vvenc/vvenc_git/source/Lib/CommonLib/arm/neon/Buffer_neon.cpp:69: numSamples must be power of two >= 4
Aborted (core dumped)
```

On x86, this seems to work (passes the unit test) since it writes beyond the length of the buffer for sizes that are not multiples of 16 (AVX2) or 8.

I have a patch coming for Arm.

micro-arm:
Not yet. I will put up a patch today. :) Just testing a few more encoding presets.

PR #563, opened 2025-06-17
Fix addAvg_neon (no stride) and its unit test to run multiples of four instead of powers-of-two

micro-arm:
Arm: Fix addAvg_neon (no stride)

- Running a real-world encode can produce non-power-of-two numSamples in the addAvg_neon (no stride) case. Update the logic to correctly handle multiples of four instead.

Fix unit test of addAvg (no stride) and disable x86 impl

- The unit test for the no-stride case only covers power-of-two heights and widths, but running a real encode shows this is insufficient. Remove these tests and instead test randomized numSamples that are multiples of four. Add four extra size to destination buffer to compare writing out of bounds.

- Disable the x86 implementation as it fails for non-power-of-two lengths in this revised unit test.

This pull request addresses issue #562.

### DIFF
--- a/source/Lib/CommonLib/x86/BufferX86.h
+++ b/source/Lib/CommonLib/x86/BufferX86.h
@@ -2552,7 +2552,8 @@ uint64_t AvgHighPassWithDownsamplingDiff2nd_SIMD (const int width,const int heig
 template<X86_VEXT vext>
 void PelBufferOps::_initPelBufOpsX86()
 {
-  addAvg   = addAvg_SSE<vext>;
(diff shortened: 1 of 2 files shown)


micro-arm:
Hello and apologies. After re-running the encode, it turns out the CHECK I put in the Arm neon is wrong and the actual sizes are correct multiples of 16 but not powers of two. So the x86 code is fine (and so is my Neon, just the CHECK is wrong).

For context, these are the widths from the real encode that triggered the CHECK.
```
numSamples=48
numSamples=192
```

I'll put up a new PR to fix the CHECK instead!

adamjw24:
Thanks for the analysis and the update!

PR #574, opened 2025-06-26
Add unit tests for applyBlock functions

athulya-arm:
Add unit tests for MCTF::applyBlock functions. Add tolerance to applyBlock comparison to allow 1 bit difference in SIMDe vs C reference.

### DIFF
--- a/test/vvenc_unit_test/vvenc_unit_test.cpp
+++ b/test/vvenc_unit_test/vvenc_unit_test.cpp
@@ -59,6 +59,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "CommonLib/RdCost.h"
 #include "CommonLib/TrQuant_EMT.h"
 #include "CommonLib/TypeDef.h"
+#include "CommonLib/Unit.h"
+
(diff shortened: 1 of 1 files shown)


adamjw24:
Why is it required to add the tolerance?

athulya-arm:
The SIMDe implementation of applyBlock may differ by one bit compared to the reference C implementation. It seems like the accuracy difference is from the floating point arithmetic.
Adjusted tolerance to reflect this.

adamjw24:
Ah true, the function has floating point strength scaling. Could you comment that its because of the floating point? Because usually we want the functions to be bit-exact

athulya-arm:
There is already a comment in the code above the compare_values_2d call about the difference seen when running the SIMDe implementation. Do you want me to add an additional comment somewhere else?

To be clear, the Neon implementation output matches with those of the scalar C implementation. The difference is observed only in the SIMDe implementation.

adamjw24:
Indeed, I missed that - sorry and thank for pointing it out.

PR #575, opened 2025-06-27
Add unit tests for TCoeffOps::cpyCoeff

micro-arm:
Add unit tests for cpyCoeff4 and cpyCoeff8.

### DIFF
--- a/test/vvenc_unit_test/vvenc_unit_test.cpp
+++ b/test/vvenc_unit_test/vvenc_unit_test.cpp
@@ -276,6 +276,75 @@ static bool test_IntraPred()
 #endif // ENABLE_SIMD_OPT_INTRAPRED
 
 #if ENABLE_SIMD_TRAFO
+static bool check_cpyCoeff( TCoeffOps* ref, TCoeffOps* opt, unsigned num_cases )
+{
+  DimensionGenerator dim;
+
+  static constexpr size_t buf_size = MAX_CU_SIZE * MAX_CU_SIZE;
+
+  // Use xMalloc to create aligned buffers for x86.
+  Pel    *src     = ( Pel*    )xMalloc( Pel,    buf_size );
+  TCoeff *dst_ref = ( TCoeff* )xMalloc( TCoeff, buf_size );
+  TCoeff *dst_opt = ( TCoeff* )xMalloc( TCoeff, buf_size );
+
+  bool passed = true;
+
+  // This function copies the prediction residue, which is signed 11-bit.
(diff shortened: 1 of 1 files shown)


test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -276,6 +276,66 @@ static bool test_IntraPred()
...
+  static constexpr size_t buf_size = MAX_CU_SIZE * MAX_CU_SIZE;
+
+  // Use xMalloc to create aligned buffers for x86.
+  Pel    *src     = ( Pel*    )xMalloc( Pel,    buf_size );
+  TCoeff *dst_ref = ( TCoeff* )xMalloc( TCoeff, buf_size );
+  TCoeff *dst_opt = ( TCoeff* )xMalloc( TCoeff, buf_size );
+
+  bool passed = true;
+
+  InputGenerator<Pel> inp_gen{ 10, /*is_signed=*/false }; // unsigned 10-bit
```
adamjw24, line 292: This function is used to copy the prediction residual, thus the value range is **signed** 10 bit (or 8-bit).

test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -276,6 +276,66 @@ static bool test_IntraPred()
...
+
+  std::ostringstream sstm_test;
+  sstm_test << "TCoeffOps::cpyCoeff" << W;
+  std::cout << "Testing " << sstm_test.str() << std::endl;
+
+  for( unsigned n = 0; n < num_cases; n++ )
+  {
+    unsigned  height    = dim.get( W, MAX_CU_SIZE );
+    unsigned  width     = dim.get( W, MAX_CU_SIZE, W ); // Width must be a multiple of W.
```
adamjw24, line 301: I think that both width and height can only be powers of 2 (including `1` and `2`, because of ISP). I guess we handle the w==1 and w==2 manually and the generalization to multiples of 4 doesnt matter. But if later on someone writes a more strict specialization it would fail the test while being correct.

micro-arm:
Hi @adamjw24,

I pushed the modified unit test for your review. Further investigation showed that the actual src values can reach beyond `signed 10-bit` [-512, 511] for a 1080p 8-bit encode so I changed the input generator to use `signed 11-bit`.

Actual values printed from an encode:
```
src_min = -752
src_max = 886
```

adamjw24:
True, a non-absolute difference of two 10-bit unsigned values is signed 11 bit :+1:

PR #576, opened 2025-06-27
Arm: Add Neon implementations of TCoeffOps::cpyCoeff kernels

micro-arm:
Add Neon implementations of cpyCoeff8 for widths that are multiples of 8, and cpyCoeff4 for widths of 4.

The calls to these functions only pass widths and heights that are powers of two.

These Neon implementations are around 1.72x faster than the SIMDe versions when benchmarked on Neoverse N/V-series microarchitectures using LLVM 20.


micro-arm:
Hi @adamjw24,

I updated this pull request to adapt to power-of-two input widths only. Therefore, `cpyCoeff8` handles only multiples of 8 while `cpyCoeff4` handles widths of 4 only.

PR #577, opened 2025-07-03
Add unit tests for TCoeffOps::roundClip

micro-arm:
Use roundClip8 for widths >= 8 and roundClip4 for width == 4.

Heights and widths are limited to powers of two:
Height: 1, 2, 4, 8, 16, 32, 64
Width: 4, 8, 16, 32, 64

----
This unit test is similar to cpyCoeff, with some added parameters (shift, round, outputMin, outputMax) where I have assigned values that come from the calling function, `TrQuant::xIT`.

The `outputMin` and `outputMax` are currently fixed to INT16 MIN/MAX range as confirmed from:
Given that `maxLog2TrDynamicRange = 15`.
https://github.com/fraunhoferhhi/vvenc/blob/5c2b2cbb5b5cda90bc05ebeb4333e73146c12525/source/Lib/CommonLib/TrQuant.cpp#L581-L582

The `shift` value range of 7 to 13 can be confirmed here:
Given that `TRANSFORM_MATRIX_SHIFT = 6` and minimum `bitDepth = 8`.
https://github.com/fraunhoferhhi/vvenc/blob/5c2b2cbb5b5cda90bc05ebeb4333e73146c12525/source/Lib/CommonLib/TrQuant.cpp#L613-L614
The `shift + 1` here makes the maximum shift value to 13, if minimum `bitDepth = 8`.
https://github.com/fraunhoferhhi/vvenc/blob/5c2b2cbb5b5cda90bc05ebeb4333e73146c12525/source/Lib/CommonLib/TrQuant.cpp#L622-L624

### DIFF
--- a/test/vvenc_unit_test/vvenc_unit_test.cpp
+++ b/test/vvenc_unit_test/vvenc_unit_test.cpp
@@ -49,7 +49,9 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <iostream>
 #include <limits.h>
 #include <stddef.h>
+#include <stdint.h>
 #include <stdio.h>
+#include <string.h>
 #include <time.h>
 
 #include "CommonLib/AffineGradientSearch.h"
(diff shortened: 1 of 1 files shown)


test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -345,6 +347,75 @@ static bool check_cpyCoeff( TCoeffOps* ref, TCoeffOps* opt, unsigned num_cases )
...
+
+  std::ostringstream sstm_test;
+  sstm_test << "TCoeffOps::roundClip";
+  std::cout << "Testing " << sstm_test.str() << std::endl;
+
+  for( unsigned n = 0; n < num_cases; n++ )
+  {
+    // Set height, width, and stride to powers of two.
+    unsigned log2height = dim.get( 0, 7 );
+    unsigned log2width = dim.get( 2, 7 );
```
adamjw24, line 373: Strictly speaking, the largest transform block in VVC is 64x64, so both W/H should only go until 64.

test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -345,6 +347,75 @@ static bool check_cpyCoeff( TCoeffOps* ref, TCoeffOps* opt, unsigned num_cases )
...
+    unsigned log2stride = dim.get( log2width, 7 );
+    unsigned height = 1 << log2height; // From 1 to 128.
+    unsigned width = 1 << log2width;   // From 4 to 128.
+    unsigned stride = 1 << log2stride; // From width to 128.
+    TCoeff shift = dim.get( 7, 13 );   // Shift range is 7 to 13.
```
adamjw24, line 378: The shift has to be such that:
input_bd - shift = output_bd (and output BD is the internal bit depth - so either 10 or 8)

adamjw24 (commented):
I think that leaving it totally to chance if roundClip4/8 are tested is not a good approach.

I would do a loop over if testing roundClip4 or 8, than a loop over internal bitdepth of 8/10 and as than a loop of number of test cases.

Please also make sure to align the bitdetph of the input generator and the shift.

Best, Adam

micro-arm:
Hi @adamjw24 ,

I have applied all changes based on your recommendations:
- Separate tests for roundClip4 and roundClip8
- Test for 8-bit / 10-bit
- Set `shift = input_bd - (output_bd + 1)`
- Set `width = 4` for W == 4
- Set `width = {8, 16, 32, 64}` for W == 8.
- I also noticed from the function call that the stride is always equal to width so I just fixed `stride = width`.

Please have a look and let me know if it needs more tweaking. Thanks.

PR #578, opened 2025-07-03
Arm: Add Neon implementations of TCoeffOps::roundClip kernels

micro-arm:
Add Neon implementations of roundClip8 for widths that are multiples of 8, and roundClip4 for widths of 4.

The calls to these functions only pass widths and heights that are powers of two.

These Neon implementations are around 1.67x faster than the SIMDe versions when benchmarked on Neoverse N/V-series microarchitectures using LLVM 20.

-----
The Neon implementations are specialized to handle the current use case, where:

* Input data are expected to be within signed 23-bit.
* Minimum right shift is 7.
* outputMax is INT16_MAX.
* outputMin is INT16_MIN.

If all of these are satisfied, the min/max clipping operations can be removed and only rounding shift is needed.

### DIFF
--- a/source/Lib/CommonLib/arm/neon/Trafo_neon.cpp
+++ b/source/Lib/CommonLib/arm/neon/Trafo_neon.cpp
@@ -53,6 +53,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "TrQuant_EMT.h"
 #include "sum_neon.h"
 
+#include <cstdint>
+
 //! \ingroup CommonLib
 //! \{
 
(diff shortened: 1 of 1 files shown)


adamjw24:
Sorry, no - pleace replace all `CHECK`s in the SIMD function with `CHECKD`

adamjw24:
Thank you. Yes, please use CHECKD. I think the encoder controls very well in which conditions the SIMD functions are called. I think the code as it is stable enough to assume we know exactly how and when the SIMD functions are called. And if implementing something new those conditions might get validated we would be able to find it easily with the debug checks.

I was thinking that probably it is inconsistent with previous SIMD, sorry about that. Its just that it only now caught my eye.

micro-arm:
Understood! I initially thought this function is an exception, so thank you for explaining. This is noted.

PR #579, opened 2025-07-07
Set -ffp-contract=off to aid cross-platform bit-reproducability

georges-arm:
Currently when running VVenC on Arm machines, the md5sum of the output video differs from x86 even when specifying `--SIMD=SCALAR`. This is due to the default `-ffp-contract` setting on GCC and LLVM enabling fused multiply-add instructions for floating-point arithmetic, which causes small bit differences due to the lack of intermediate rounding in these instructions.

Setting `-ffp-contract=off` disables this behaviour, ensuring the md5sum output for the tested video now matches for `--SIMD=SCALAR`. There are some remaining differences when other SIMD settings are used on Arm which will be resolved in a separate PR.


georges-arm:
> Do you know what the impact is on performance? I think we would prefer performance over bit-equality (as the bitstreams are all VVC compliant and the only differences might be different decisions due to rounding errors in cost calculation, so no big BD-rate impact).

Performance impact seems to be about a ~0.3% slowdown on a whole-video encoding, benchmarked on both an Arm and x86 machine.

georges-arm:
> Could this be guarded by an additional CMake option? I.e. `VVENC_FPP_CONTRACT_OFF` - so that it can be decided at build if to prioritize speed or bit-equality

Sure! Any preference on a default setting?

PR #580, opened 2025-07-17
Add missing "using namespace arm_simd" in IntraPrediction.h

georges-arm:
The `IntraPrediction.h` includes uses of the `ARM_SIMD` type but does not import the namespace that defines these types like the other headers do. This previously worked by virtue of other includes in the relevant .cpp files aliasing the namespace ahead of this header being included, however this fails when selectively disabling some SIMD implementations.

To fix this we can simply enable the namespace import, copying the style that is already used in the other headers.


adamjw24:
Thank you!

PR #598, opened 2025-08-11
Arm: Improve Neon implementations of InterpolationFilter::filterWxH_N8

micro-arm:
Arm: Change Neon implementation of simdFilter4xH_N8 to simdFilter4x4_N6
- The 8-tap filter for filter4x4 can be reduced to a 6-tap filter due to the zero coefficients at the 0th and 7th indices.
- Unroll the horizontal filter calculation from the do-while loop to eliminate the loop-carried dependency of copying results to the next iteration, and fix the height loop to 4, as this is the only use case for this function.
- These changes improve the performance by 34% compared to the previous 8-tap Neon implementation when benchmarked on a Neoverse V2 micro-architecture using LLVM 20.

These changes are applied to all width (4x4, 8xH, 16xH) implementations:
- Replace the max calculation for the headRoom with a CHECKD instead to ensure headRoom >= 2.
- Replace vqmovn with vqmovun for isLast == true case to save a vdup + vmax instruction.
- Also, rename helper functions to improve readability.

### DIFF
--- a/source/Lib/CommonLib/arm/neon/InterpolationFilter_neon.cpp
+++ b/source/Lib/CommonLib/arm/neon/InterpolationFilter_neon.cpp
@@ -170,7 +170,7 @@ static inline int16x4_t filter_horiz_4x1_N8_neon( Pel const* src, int16x8_t ch, 
   return vqmovn_s32( vsuma );
 }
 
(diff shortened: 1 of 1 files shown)


adamjw24:
Thank you!

PR #603, opened 2025-08-13
RateCtrl.cpp: Fix abs bug error to add std

MartinEesmaa:
This will replace abs by std::abs to compile successfully using clang with osxcross.

- Martin Eesmaa


adamjw24:
Thank you!

PR #605, opened 2025-08-26
Add unit tests for applyFrac6Tap functions

athulya-arm:
- Add unit tests for `MCTF::applyFrac6Tap_8x` and `MCTF::applyFrac6Tap_4x` functions.
    
- Move filter tables `m_interpolationFilter4` and `m_interpolationFilter8` from the private to the public section of the 
MCTF class to allow external access from unit tests.

### DIFF
--- a/source/Lib/CommonLib/MCTF.h
+++ b/source/Lib/CommonLib/MCTF.h
@@ -154,6 +154,9 @@ private:
 #endif
 
 public:
+  static const int16_t m_interpolationFilter4[16][4];
+  static const int16_t m_interpolationFilter8[16][8];
+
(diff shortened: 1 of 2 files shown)


adamjw24:
Thanks for the code. Since you are already on it, and the change is very minor, could you add the test for the 4-tap filter as well? Just so we have more coverage (basically the only change would be to call `MCTF::m_applyFrac[c][1]` instead of `MCTF::m_applyFrac[c][0]` and use `MCTF::m_interpolationFilter4` instead of `MCTF::m_interpolationFilter8` as filter inputs.

adamjw24:
> And of course one more loop is needed...

Or you add `bool is4tap` as paramter to `check_one_applyFrac6Tap` (which would of course not be named 6Tap anymore) and call it twice.

adamjw24:
Thanks!

PR #609, opened 2025-09-16
sub_SSE: Use stores that don't require aligned access

micro-arm:
The W==8 path for `sub_SSE` uses `_mm_store_si128`, which requires 16-byte pointer alignment of the store address.

When building for 32-bit Arm platforms using the x86 intrinsics through SIMDe, for some versions of LLVM this is emitted as a store instruction with a 16-byte alignment hint. At runtime the pointer is observed to only have 8-byte alignment, causing an alignment fault on 32-bit Arm platforms with alignment checks enabled.

To work around this, switch to `_mm_storeu_si128` which does not have the alignment requirement, allowing the unit tests on 32-bit Arm platforms to pass successfully.


adamjw24:
Good catch - I'm surprised it didnt break anything on Linux for x86. Thanks!

PR #613, opened 2025-09-26
Arm: Update motionErrorLumaFrac_loRes_neon

athulya-arm:
Update Neon implementation of MCTF motionErrorLumaFrac_loRes. 
This improved implementation is approximately 36% faster than the existing Neon version.

### DIFF
--- a/source/Lib/CommonLib/arm/neon/MCTF_neon.cpp
+++ b/source/Lib/CommonLib/arm/neon/MCTF_neon.cpp
@@ -64,92 +64,247 @@ POSSIBILITY OF SUCH DAMAGE.
 namespace vvenc
 {
 
-static int16x8_t motionErrorLumaFrac_loRes_step( const int16x8_t xf, const Pel* rowStart, const Pel maxSampleValue )
+enum class FilterCoeffType4
 {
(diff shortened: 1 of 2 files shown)


adamjw24:
How much influence does the symmetry optimization bring?

Thats an interesting approach, that could maybe be borrowed to other to other modules and X86.

athulya-arm:
On average, `FullSymmetric` is ~3.3% faster than `Generic` and `SkewLeft/Right` is ~11% faster than `Generic`, at the cost of a slight increase in code memory.
In sample video runs, for the function, `Generic` accounts for ~80% of occurrences, while `SkewLeft/Right` and `FullSymmetric` together make up the remaining ~20%.

adamjw24:
> On average, `FullSymmetric` is ~3.3% faster than `Generic` and `SkewLeft/Right` is ~11% faster than `Generic`, at the cost of a slight increase in code memory. In sample video runs, for the function, `Generic` accounts for ~80% of occurrences, while `SkewLeft/Right` and `FullSymmetric` together make up the remaining ~20%.

Thanks! So the gains are marginal, right?

PR #615, opened 2025-09-29
simdInterpolateVerM1: Fix the 8-tap coeff load in x86

micro-arm:
The path for `N == 8` was incorrectly loading filter taps using `_mm_set1_epi64x(*(int64_t*)coeff)`, which broadcasts only the low 64 bits. That duplicated coeff[0..3] into both halves of the vector, so the dot product used [c0 c1 c2 c3 c0 c1 c2 c3] instead of the intended [c0 c1 c2 c3 c4 c5 c6 c7].

Fix the N8 path by loading all eight int16_t coefficients instead.


adamjw24:
Good catch! Now I understand why the option `ReduceFilterME=2` (using 4-tap for ME) works so well. It is because when using 6- or 8-tap ME, the last row will be all broken...

PR #616, opened 2025-09-30
simdInterpolateHorM1: Fix the tail path to 16-bit clipping

micro-arm:
This makes the tail behavior consistent with the main loop and fixes incorrect results on tiny blocks (e.g. w=1, h=1) when clip is set.


adamjw24:
It's a bigger fix. I'm looking into it right now...

PR #622, opened 2025-10-09
Arm: Enable SVE intrinsics by default if compiler supports it

georges-arm:
In #443 we added support for SVE and SVE2 feature detection. This was initially disabled by default as there were no optimized implementations using SVE at the time.

Now that we have quite a few SVE optimized functions, enable SVE by default if the compiler supports it. SVE2 is left disabled by default since there is still no code making use of it yet.

Benchmarking on a Neoverse V2 machine with LLVM 20, this improves performance of a --preset=medium encoding by ~3.7% over the previous default configuration.


K-os:
Oh, ok I get it. I think then it makes sense to leave SVE2 disabled for now to prevent confusion, and enable it once it's implemented.

PR #645, opened 2026-02-08
make full static build on Linux , github actions Build.yml add upload-artifact

WLaoDuo:
Update CMakeList.txt to make  full static build on Linux .The old static build version still needs *.so to run on other system , this new version don't need any dependencies.

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -85,12 +85,3 @@ jobs:
         cmake --build build/static --config Release
         cmake -S . -B build/shared  -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=1 -A "${{ matrix.config.msvc_arch }}"
         cmake --build build/shared --config Release
-    
-    - name: Upload Artifacts
-      uses: actions/upload-artifact@v4
-      with:
-        name: ${{ matrix.config.name }}-artifacts
-        path: |
-          bin/
-          lib/
-        if-no-files-found: warn


WLaoDuo:
> Hi WLaoDuo, Thanks for the suggestion, but we will not merge this. Fully static linking of the executable is not necessarily what you want, when just building the static VVenC library.
> 
> We could add another configuration option, to build a fully static binary, but I would prefer not to. I think if you really want to have a static binary, you can still always pass `-DCMAKE_CXX_FLAGS=-static` instead of something like `-DVVENC_FULLY_STATIC_BUILD=ON`, especially if you want to add other niche options like `-no-pie`.

Hello , @K-os 
Thanks for your answer, I tried `-DCMAKE_CXX_FLAGS=-static` on alpine 3.23.2 , `interfacetest` can't be built , but when I add `CMAKE_CXX_FLAGS=-static -no-pie` to CMakeList.txt , it can be built. If you don't need this PR, I will close it soon.

K-os:
Ok, that seems to be an alpine specific problem, which I don't think needs to be addressed in the vvdec build system.

Alpine uses a different libc implementation than most other linux distributions, seems not to support statically linked position independent executables.

K-os:
I think I found the solution: for me building of the interfacetest had nothing to do with `-pie`/`-no-pie`, but since it is a C binary, it also needs `-DCMAKE_C_FLAGS=-static` to build.

PR #655, opened 2026-02-26
Fix parity tracking in DQ SIMD

### LINKED ISSUE #654: DepQuant - Differences observed in X86/SIMDe code path
We're investigating porting the DepQuant code to a native Neon implementation, and reviewed the current x86/SIMDe code paths against the scalar C reference for
- State::updateStates
- State::updateStatesEOS

We identified some differences caused by the use of signed 8-bit packing and arithmetic in the x86 path, which do not match the intended scalar behaviour.

### Issue in State::updateStates
**Current x86 Code**
```
__m128i a_m = _mm_min_epi16( v, _mm_add_epi16( v254_4, a_1 ) ); 
 a_m = _mm_packs_epi16( a_m, vn1 );
```
`_mm_packs_epi16` performs _signed saturation, values above 127 gets clipped, even though we just clipped at 254/255 at the above step. That completely overrides the 254/255 clipping done at the previous stage.

Similar logic in scalar C is as below
`m_sbb.absLevels[scanInfo.insidePos] = ( uint8_t ) std::min<TCoeff>( 254 + ( decision.absLevel & 1 ), decision.absLevel ); 
`

**Proposed changes**

1. Replace _mm_packs_epi16
We could use _mm_packus_epi16 instead (as below) which would pack and narrow while maintaining values in the [0..255] range.
`a_m = _mm_packus_epi16(a_m, _mm_setzero_si128()); 
`
2. Change int8_t l[4] to uint8_t l[4], since we now treat them as unsigned bytes. All related loads (and store) needs change from signed to unsigned.
3. Update all subsequent operations that uses those arrays to use the unsigned byte intrinsics
- replace _mm_sign_epi8 trick (breaks for bytes >=128)
`   a_m = _mm_or_si128( a_m, _mm_sign_epi8( v032, a_m ) ); 
`
- numSig decrement
```
__m128i lvls = _mm_loadu_si32( l );
lvls = _mm_cmpgt_epi8( lvls, _mm_setzero_si128() );
numSig = _mm_subs_epi8( numSig, lvls );   
```
   - Bin index clamp
```
__m128i mlvl = _mm_loadu_si32( l );    
__m128i mbins = _mm_min_epi8( mlvl, _mm_set1_epi8( 2 ) );   
```
   - update_deps_vec
  ```
 __m128i msum = _mm_loadu_si32( &curr.sum1st[addr] ); 
   msum = _mm_adds_epu8( msum, mlvl );  
```

### Issue in State::updateStatesEOS
Similar issue exists here.

**Current x86 Code**
```
int8_t s[4] = { 0 }, l[4] = { 0 }, z[4] = { 0 };  

.....

l[i] = s[i] > -2 ? std::min<int>( decisions.absLevel[i], 254 + ( decisions.absLevel[i] & 1 ) ) : 0;
```

The l[4] array is declared as int8_t (decisions.absLevel is of type int16_t), causing capped values above 127 to overflow and wrap into negatives.

adamjw24:
The `_mm_packs_epi16` clamps to int8 range. While only the absolute value up to 51 is really needed to be tracked, it is necessary to properly track the parity (for proper context selection). This fixes the range during clamping.

C code changes are not necessary, but applied to improve readability.

Fixes #654 

@athulya-arm could you confirm the issues is herewith fixed?

### DIFF
--- a/source/Lib/CommonLib/DepQuant.cpp
+++ b/source/Lib/CommonLib/DepQuant.cpp
@@ -931,7 +931,7 @@ namespace DQIntern
 
       if( decision.absLevel )
       {
-        m_sbb.absLevels[scanInfo.insidePos] = ( uint8_t ) std::min<TCoeff>( 254 + ( decision.absLevel & 1 ), decision.absLevel );
+        m_sbb.absLevels[scanInfo.insidePos] = ( uint8_t ) std::min<TCoeff>( 126 + ( decision.absLevel & 1 ), decision.absLevel );
         
         if( scanInfo.currNbInfoSbb.numInv )
         {
@@ -1024,7 +1024,7 @@ namespace DQIntern
         ::memset( m_sbb.absLevels, 0, sizeof( m_sbb.absLevels ) );
       }
 
-      m_sbb.absLevels[ scanInfo.insidePos ] = (uint8_t)std::min<TCoeff>( 254 + ( decision.absLevel & 1 ), decision.absLevel );
+      m_sbb.absLevels[ scanInfo.insidePos ] = (uint8_t)std::min<TCoeff>( 126 + ( decision.absLevel & 1 ), decision.absLevel );
 
       m_commonCtx.update( scanInfo, prvState, *this );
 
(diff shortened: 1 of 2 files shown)


athulya-arm:
Yes, all reported issues have been addressed. Thank you.

PR #684, opened 2026-04-02
thirdparty: update bundled SIMDE to fix link error on musl/non-glibc

kjg0724:
Reopens #682.

Accidentally closed the previous PR — apologies for the confusion.

Updates thirdparty/simde to simd-everywhere/simde@cb5b957, which includes:
- roundeven/roundevenf guard (simd-everywhere/simde#1398)
- exp10/exp10f guard (simd-everywhere/simde#1400)

Both fixes address link errors on musl/non-glibc systems. CI was green on #682.


K-os (changes requested):
Thanks, in principle this looks good. I just have one small change request:
We pull the simd-everywhere dependency directly from this repository: https://github.com/simd-everywhere/simde-no-tests.git
So, could you please update the commit message to point to the corresponding commit in that repository to prevent confusion? I think it should be simd-everywhere/simde-no-tests@8574c41

kjg0724:
Updated the commit message to reference simd-everywhere/simde-no-tests@8574c41 as requested.

K-os (approved):
Ok, thanks. Looks good now.

PR #686, opened 2026-04-08
Add unit tests for DepQuant functions and refactor findFirstPos

athulya-arm:
- Add unit test for DepQuant functions `State::updateStates`, `State::updateStatesEOS`, `State::checkAllRdCostsOdd1`, `findFirstPos`.
- Modify the DepQuant constructor to include a new `enableOpt` parameter to create a reference object and an optimized object for testing.
- Move `m_updateStates`, `m_updateStatesEOS`, `m_checkAllRdCostsOdd1`, `m_findFirstPos` from the private to the public section of the class to enable unit test access.
- Add /wd4100 to unit test MSVC flags to suppress unreferenced parameter warnings from Quant.h, matching the existing library build configuration.
- Extract the scalar findFirstPos logic from `xQuantDQ` into a dedicated function `findFirstPos`.
- Call `m_findFirstPos` unconditionally from `xQuantDQ` and move the final coefficient-wise refinement from `xQuantDQ` into `findFirstPos` for both the C and x86 implementations.

### DIFF
--- a/source/Lib/CommonLib/DepQuant.cpp
+++ b/source/Lib/CommonLib/DepQuant.cpp
@@ -1410,7 +1410,7 @@ void DepQuant::xDequantDQ( const TransformUnit& tu,  CoeffBuf& recCoeff, const C
   m_quant.dequantBlock( tu, compID, cQP, recCoeff, enableScalingLists, piDequantCoef );
 }
 
-DepQuant::DepQuant( const Quant* other, bool enc, bool useScalingLists ) : QuantRDOQ2( other, useScalingLists ), RateEstimator(), m_commonCtx()
+DepQuant::DepQuant( const Quant* other, bool enc, bool useScalingLists, bool enableOpt ) : QuantRDOQ2( other, useScalingLists ), RateEstimator(), m_commonCtx()
 {
   const DepQuant* dq = dynamic_cast<const DepQuant*>( other );
   CHECK( other && !dq, "The DepQuant cast must be successfull!" );
@@ -1436,9 +1436,12 @@ DepQuant::DepQuant( const Quant* other, bool enc, bool useScalingLists ) : Quant
   m_updateStates        = DQIntern::updateStates;
   m_findFirstPos        = nullptr;
 
+  if( enableOpt )
+  {
 #if defined( TARGET_SIMD_X86 ) && ENABLE_SIMD_OPT_QUANT
-  initDepQuantX86();
+    initDepQuantX86();
 #endif
+  }
 }
 
 DepQuant::~DepQuant()
--- a/source/Lib/CommonLib/DepQuant.cpp
+++ b/source/Lib/CommonLib/DepQuant.cpp
@@ -55,6 +55,23 @@ namespace vvenc {
 
 namespace DQIntern
 {
+  static void findFirstPos( int& firstTestPos, const TCoeff* tCoeff, const DQIntern::TUParameters& tuPars, int defaultTh,
+                            bool zeroOutForThres, int zeroOutWidth, int zeroOutHeight )
+  {
+    for( ; firstTestPos >= 0; firstTestPos-- )
+    {
+      if( zeroOutForThres && ( tuPars.m_scanId2BlkPos[firstTestPos].x >= zeroOutWidth ||
+                              tuPars.m_scanId2BlkPos[firstTestPos].y >= zeroOutHeight ) )
+      {
+        continue;
+      }
+      if( abs( tCoeff[tuPars.m_scanId2BlkPos[firstTestPos].idx] ) > defaultTh )
+      {
+        break;
+      }
+    }
+  }
+
   void Rom::xInitScanArrays()
   {
     if( m_scansInitialized )
@@ -1169,16 +1186,7 @@ void DepQuant::xQuantDQ( TransformUnit& tu, const CCoeffBuf& srcCoeff, const Com
   {
     const TCoeff defaultTh = TCoeff( thres / ( defaultQuantisationCoefficient << 2 ) );
 
-    if( m_findFirstPos )
-    {
-      m_findFirstPos( firstTestPos, tCoeff, tuPars, defaultTh, zeroOutforThres, zeroOutWidth, zeroOutHeight );
-    }
-
-    for( ; firstTestPos >= 0; firstTestPos-- )
-    {
-      if( zeroOutforThres && ( tuPars.m_scanId2BlkPos[firstTestPos].x >= zeroOutWidth || tuPars.m_scanId2BlkPos[firstTestPos].y >= zeroOutHeight ) ) continue;
-      if( abs( tCoeff[tuPars.m_scanId2BlkPos[firstTestPos].idx] ) > defaultTh ) break;
-    }
+    m_findFirstPos( firstTestPos, tCoeff, tuPars, defaultTh, zeroOutforThres, zeroOutWidth, zeroOutHeight );
   }
 
   if( firstTestPos < 0 )
@@ -1434,7 +1442,7 @@ DepQuant::DepQuant( const Quant* other, bool enc, bool useScalingLists, bool ena
   m_checkAllRdCostsOdd1 = DQIntern::checkAllRdCostsOdd1;
   m_updateStatesEOS     = DQIntern::updateStatesEOS;
   m_updateStates        = DQIntern::updateStates;
-  m_findFirstPos        = nullptr;
+  m_findFirstPos        = DQIntern::findFirstPos;
 
   if( enableOpt )
   {
--- a/test/vvenc_unit_test/vvenc_unit_test.cpp
+++ b/test/vvenc_unit_test/vvenc_unit_test.cpp
@@ -372,6 +372,159 @@ static bool check_updateStates( DepQuant* ref, DepQuant* opt, unsigned num_cases
   return passed;
 }
 
+static bool check_updateStatesEOS( DepQuant* ref, DepQuant* opt, unsigned num_cases )
+{
+  printf( "Testing DepQuant::updateStatesEOS\n" );
+
+  InputGenerator<TCoeff> g4{ 4, /*is_signed=*/false };
+  InputGenerator<TCoeff> g6{ 6, /*is_signed=*/false };
+  InputGenerator<TCoeff> g8{ 8, /*is_signed=*/false };
+  InputGenerator<TCoeff> g11{ 11 };
+  InputGenerator<TCoeff> g31{ 31 };
+  DimensionGenerator rng;
+
+  std::ostringstream sstm;
+  sstm << "DepQuant updateStatesEOS";
+
+  DQIntern::ScanInfo scanInfo;
+  DQIntern::Decisions decisions;
+  DQIntern::StateMem skip;
+  DQIntern::StateMem curr_ref;
+  DQIntern::StateMem curr_opt;
+  DQIntern::CommonCtx commonCtx_ref;
+  DQIntern::CommonCtx commonCtx_opt;
+
+  DQIntern::Rom rom;
+  rom.init();
+  DQIntern::TUParameters tuPars( rom, 16, 16, CH_C );
+
+  FracBitsAccess fracBitsAccess{ 1 };
+  CodingUnit cu;
+  TransformUnit tu;
+  tu.cu = &cu;
+
+  DQIntern::RateEstimator rateEst;
+  rateEst.initCtx( tuPars, tu, COMP_Cb, fracBitsAccess );
+
+  commonCtx_ref.reset( tuPars, rateEst );
+  commonCtx_ref.swap();
+  commonCtx_ref.reset( tuPars, rateEst ); // Rebinds m_currSbbCtx/m_prevSbbCtx entries to parts of m_memory.
+
+  commonCtx_opt.reset( tuPars, rateEst );
+  commonCtx_opt.swap();
+  commonCtx_opt.reset( tuPars, rateEst );
+
+  bool passed = true;
+
+  for( unsigned i = 0; i < num_cases; ++i )
+  {
+    std::memset( &curr_ref, 0xCD, sizeof( curr_ref ) );
+    std::memset( &curr_opt, 0xCD, sizeof( curr_opt ) );
+
+    scanInfo.scanIdx = rng.get( 16, 128, 16 );
+    scanInfo.sbbSize = rng.getOneOf<int8_t>( { 4, 16 } );
+    scanInfo.numSbb = 16;
+
+    scanInfo.sbbPos = rng.get( 0, scanInfo.numSbb - 1 );
+    scanInfo.nextSbbRight = 0;
+    scanInfo.nextSbbBelow = 0;
+
+    scanInfo.insidePos = 0;
+    scanInfo.nextInsidePos = 15;
+    scanInfo.gtxCtxOffsetNext = ( int8_t )rng.get( 0, 200 );
+    scanInfo.sigCtxOffsetNext = ( int8_t )rng.get( 0, 200 );
+    scanInfo.currNbInfoSbb.numInv = 0;
+    std::generate( scanInfo.currNbInfoSbb.invInPos, scanInfo.currNbInfoSbb.invInPos + 5, g4 );
+
+    for( int idx = 0; idx < 4; ++idx )
+    {
+      // prevId >= 4 implies prevId - 4 == idx.
+      decisions.prevId[idx] = rng.getOneOf<int8_t>( { -2, -1, 0, 1, 2, 3, int8_t( idx + 4 ) } );
+
+      curr_ref.refSbbCtxId[idx] = rng.getOneOf<int8_t>( { -1, 0, 1, 2, 3 } );
+      if( decisions.prevId[idx] >= 4 )
+      {
+        decisions.absLevel[idx] = 0;
+      }
+      else
+      {
+        // prevId == -1 indicates this would be the first non-zero coefficient (thus absVal > 0).
+        int minLevel = decisions.prevId[idx] >= 0 ? 0 : 1;
+        decisions.absLevel[idx] = ( TCoeffSig )rng.get( minLevel, 127 );
+      }
+    }
+    std::generate( decisions.rdCost, decisions.rdCost + 4, g31 );
+
+    curr_ref.initRemRegBins = 4;
+    std::generate( curr_ref.numSig, curr_ref.numSig + 4, g6 );
+    std::generate( curr_ref.remRegBins, curr_ref.remRegBins + 4, g11 );
+    for( int idx = 0; idx < 16; ++idx )
+    {
+      std::generate( curr_ref.tplAcc[idx], curr_ref.tplAcc[idx] + 4, g8 );
+      std::generate( curr_ref.sum1st[idx], curr_ref.sum1st[idx] + 4, g8 );
+    }
+
+    std::generate( skip.remRegBins, skip.remRegBins + 4, g11 );
+
+    std::fill_n( &curr_ref.absVal[0][0], 16 * 4, 0 ); // AbsVal is always set to 0 prior to updateStates.
+
+    curr_opt = curr_ref;
+
+    ref->m_updateStatesEOS( scanInfo, decisions, skip, curr_ref, commonCtx_ref );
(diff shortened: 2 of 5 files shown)


adamjw24:
If you are adding a reference C implementation of `findFirstPos`, maybe we should just use that? What I mean is that the virtual implementation is supposed to find SBB-exact pointer to the first codec coefficient and than `xQuantDq` refines that to coefficient exact resolution. If we have a C reference implementation of `findFirstPos`, maybe it would be just easier to say the function returns a cofficient exact first pos. This would require to put the refinement at the end of the SIMD as well as C implementations, and to remove it from `xQuantDq`. The nullptr-check if `xQuantDq` could also be removed.

source/Lib/CommonLib/DepQuant.cpp
```diff
@@ -55,6 +55,69 @@ namespace vvenc {
...
+          break;
+        }
+      }
+      if( firstTestPos >= 0 )
+      {
+        // If a coefficient was found, advance the pointer to the end of the current subblock
+        // for the subsequent coefficient-wise refinement.
+        firstTestPos += sbbSize - 1;
+      }
+    }
```
adamjw24: I think all of this is obsolete, no?

athulya-arm: Do you mean the whole function or the newly added logic inside it?
We meant adding a scalar reference C would be useful to serve as a generic fallback, a readable reference for future SIMD implementations like Neon, rather than relying on the x86 version to understand the logic and also as a baseline for unit tests to compare optimized implementations against.

adamjw24: The newly added logic - I did ask for the function previously myself.

I guess you are right, its not necessary but doesn't hurt either. Could you maybe add a comment that its just a reference for SIMD and that the following loop would also take care of that?

athulya-arm: Thanks, I removed the scalar 4x4 prescan from `findFirstPos` to avoid confusion. Also updated the unit test for "not-found cases" to check for a negative pos, which is all `xQuantDQ` depends on.

test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -253,6 +254,417 @@ class DimensionGenerator
...
+        state.ctx.cff[j] = rng.get( state.cffBitsCtxOffset, state.cffBitsCtxOffset + 4 );
+
+        for( int k = 0; k < DQIntern::RateEstimator::sm_maxNumSigCtx; ++k )
+        {
+          sigBuf[j][k].intBits[0] = g20();
+          sigBuf[j][k].intBits[1] = g20();
+        }
+        decisions_ref.rdCost[j] = decisions_opt.rdCost[j] = DQIntern::rdCostInit >> 2;
+        decisions_ref.absLevel[j] = decisions_opt.absLevel[j] = -1;
+        decisions_ref.prevId[j] = decisions_opt.prevId[j] = -2;
```
adamjw24: `m_checkAllRdCostsOdd1` if the trellis has already been started, i.e. prevId != -2.

The only function that can correctly start a trellis path is `checkRdCostStart`. Sorry about that, this one you really just have to know....

The easy solution would be to just initialize prevId (at least for some states) to [0...3] and absLevel to a positive integer.

Also, an interesting test would be to initlize state 0 and 2 to prevId [0...3] prevId and positive absLevel and have states 1 and 3 initialized as you did - this way correct handling of mixed valid and invalid paths can be tested.

test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -253,6 +254,417 @@ class DimensionGenerator
...
+    scanInfo.insidePos = rng.get( 0, 15 );
+    scanInfo.nextInsidePos = rng.get( 0, 15 );
+    scanInfo.gtxCtxOffsetNext = ( int8_t )rng.get( 0, 200 );
+    scanInfo.sigCtxOffsetNext = ( int8_t )rng.get( 0, 200 );
+    scanInfo.currNbInfoSbb.numInv = ( uint8_t )rng.get( 0, 5 );
+    std::generate( scanInfo.currNbInfoSbb.invInPos, scanInfo.currNbInfoSbb.invInPos + 5, g4 );
+
+    for( int idx = 0; idx < 4; ++idx )
+    {
+      decisions.prevId[idx] = rng.getOneOf<int8_t>( { -1, 0, 1, 2, 3 } );
```
adamjw24: Keep testing -2, so that we also keep track of invalid (or - not yet started), trellis paths.

test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -253,6 +254,417 @@ class DimensionGenerator
...
+
+  bool passed = true;
+
+  for( unsigned i = 0; i < num_cases; ++i )
+  {
+    std::memset( &curr_ref, 0xCD, sizeof( curr_ref ) );
+    std::memset( &curr_opt, 0xCD, sizeof( curr_opt ) );
+
+    scanInfo.scanIdx = rng.get( 16, 128, 16 );
+    scanInfo.sbbSize = rng.getOneOf<int8_t>( { 4, 8, 16 } );
```
adamjw24: Does sbbSize of 8 (i.e. 4x2 or 2x4) exist? I though its either 2x2 or 4x4 for coefficient subblocks.

adamjw24:
Thank you for including the tests. This is very low level code so its full of preassumptions that are only clear if you have in-depth knowledge of the VVC standard.

I have some general comments here. Using random number generator to generate the rdCosts for the `checkRdCost*` functions might make the values so far apart, that it would not show if decisions were based on faulty calcualtions (just because the values are not close-by, and the output is always only the best found cost - the dismissed options are not cross-checked).

Further, there are really many cases that can, and eventually will, arise. Here, I would propose a different testing scheme, more like integration tests. I.e. generate different "realistic" residual signals (e.g. by generating random or even non-random signals, than filtering them somehow, and using the difference as residual), applying the forward transform, quantization, and than checking the resulting transform coefficients. Do that at a wide range of QPs (e.g. 0-50 in steps of x) for a wide range of blocks sizes and transforms (to test zero-out handling) and most cases should be tested. The biggest issue here would be generating apropriate residuals.

If the current way of testing (unit) works better for you, I guess thats also ok, but I see many cases that are not caught with the approach, as commented above. Once you have some implementation I would also test some corpus of sequences at wide range of QPs versus the scalar implementation and see if the bitstreams are bit-equal. This is ultimately what we did when developing the x86 simd. Anyway, I am ok to proceed as is, but would be great if at least the issues mentioned were resolved.

athulya-arm:
Thank you for the feedback. For this patch, we plan to strengthen the existing unit test coverage rather than adding integration style transform+quant tests. The current unit tests were already useful in bringing up the Neon implementation and we will extend them with cases like
1. tie/near-tie cases with close state.rdCost values and small pq_a_dist / pq_b_dist deltas, so a small arithmetic difference can change the selected path.
2. SCAN_EOCSBB cases with state.numSig == 0
We will also address the other review comments.

Also, following the suggestion to add testing prevId == -2 in check_updateStates, I noticed that the C and x86 paths behave differently for that case. The C path bypasses the entire update for that state, while the x86 path still updates/clears some fields. Since this case fails the comparison, I think it would be good to clarify whether this is the intended handling for an invalid path.

adamjw24:
> Also, following the suggestion to add testing prevId == -2 in check_updateStates, I noticed that the C and x86 paths behave differently for that case. The C path bypasses the entire update for that state, while the x86 path still updates/clears some fields. Since this case fails the comparison, I think it would be good to clarify whether this is the intended handling for an invalid path.

Thats what I mean when mentioning integration tests. The update in x86 for prevId=-2 are irrelevant, as long as they are correctly overwritten later. The values are only updated because of how SIMD works, and if the values written back are not reset to the inital state, its only because its irrelevant for those values.

Please give me more info on the problematic values than we can discuss the best way to resolve that.

athulya-arm:
C: https://github.com/fraunhoferhhi/vvenc/blob/master/source/Lib/CommonLib/DepQuant.cpp#L893
X86: https://github.com/fraunhoferhhi/vvenc/blob/master/source/Lib/CommonLib/x86/DepQuantX86.h#L72

For prevId == -2, the x86 path processes 4 lanes (stateId) together and still writes the corresponding lane for the invalid state, looks like a reset, while the C path bypasses the update for that state.
Pseudo-code of the resulting x86 state for that lane:
```
curr.numSig[stateId] = 0;
curr.refSbbCtxId[stateId] = -1;
curr.remRegBins[stateId] = curr.initRemRegBins;
curr.tplAcc[0..15][stateId] = 0;
curr.sum1st[0..15][stateId] = 0;
curr.ctx.cff[stateId] = scanInfo.gtxCtxOffsetNext;
curr.ctx.sig[stateId] = scanInfo.sigCtxOffsetNext;
```
If those values are indeed don’t-care for prevId == -2 and are always overwritten before becoming relevant, I can adjust the unit test accordingly.

adamjw24:
> C: https://github.com/fraunhoferhhi/vvenc/blob/master/source/Lib/CommonLib/DepQuant.cpp#L893 X86: https://github.com/fraunhoferhhi/vvenc/blob/master/source/Lib/CommonLib/x86/DepQuantX86.h#L72
> 
> For prevId == -2, the x86 path processes 4 lanes (stateId) together and still writes the corresponding lane for the invalid state, looks like a reset, while the C path bypasses the update for that state. Pseudo-code of the resulting x86 state for that lane:
> 
> ```
> curr.numSig[stateId] = 0;
> curr.refSbbCtxId[stateId] = -1;
> curr.remRegBins[stateId] = curr.initRemRegBins;
> curr.tplAcc[0..15][stateId] = 0;
> curr.sum1st[0..15][stateId] = 0;
> curr.ctx.cff[stateId] = scanInfo.gtxCtxOffsetNext;
> curr.ctx.sig[stateId] = scanInfo.sigCtxOffsetNext;
> ```
> 
> If those values are indeed don’t-care for prevId == -2 and are always overwritten before becoming relevant, I can adjust the unit test accordingly.

- `numSig` ... is overwritten with the same value (`0`)
- `refSbbCtxId` ... is overwritten with same value (`-1`)
- `sum1st` ... is overwritten with the same value (`0`)
- `remRegBins` ... could already be initalized to `initRemRegBins` in the init stage (I guess its really only not because the value is initalized after `DQInter::initStates` call). Anyway, the first valid coefficient (`prevId == -1`) will correctly overwrite it anyway
- `tplAcc` ... will be correctly fully overwritten the first time a valid coefficient is processed (`prevId == -1`)
- `ctx.cff` ... is not accessed by `checkRdCostStart`, the only function that can produce valid decision for a `prevId == -2` state
- `ctx.sig` .... the same as `ctx.cff`

In that sense, changes `remRegBins`, `tplAcc`, `ctx.cff` and `ctx.sig` for invalid paths (`prevId == -2`) are irrelevant and don't need to be compared for any invalid paths.

The most important aspect is that the `rdCost` of a `prevId == -2` decision always has to be larger than any of the valid decisions (`prevId > -2`), so that for the next coefficients its never chosen as previous `stateId`, the same as with the state for which `prevId == -2` was chosen. So:
- after `checkRdCost*` the `Decisions::rdCost[i]` for each `i` where `prevId == -2` has to be larger than for any `i` for which  for which `prevId > -2`
- after `updateState{,EOS}`, the `StateMem::rdCost[i]` for each `i` where `Decisions::prevId[i] == -2` has to be larger than any for `i` for which which `prevId > -2`

adamjw24:
Also, another aspect that might be important. `Decisions::prevId[i] == -2` might only occur at the beginning of the search when the `StateMem` is initialized with default values. Once any non-default values has been written to any of the states `Decisions::prevId[i] == -2` cannot occur. Its because for the first coefficients only transitions to states 0 and 2 are possible (because the trellis startes in state 0), thus states 1 and 3 are still invalid. But for the next coefficient, as we have valid values in previous states 0 and 2, we have valid transitions to any of the states 0...3 available.

test/vvenc_unit_test/vvenc_unit_test.cpp
athulya-arm: `absLevel` and `prevId` in `checkAllRdCostsOdd1()` are output decision values, the function does not use them as inputs. In this test, we updated their initialization to invalid values so the unit test fails by default if either the reference or optimized implementation does not overwrite them correctly. The goal here is to verify that both implementations update these decision outputs consistently, rather than to model a pre-started trellis state.

test/vvenc_unit_test/vvenc_unit_test.cpp
athulya-arm: Done

test/vvenc_unit_test/vvenc_unit_test.cpp
athulya-arm: Done

adamjw24:
Thank you for addressing the comments, I think its good to go now.

test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -253,6 +254,518 @@ class DimensionGenerator
...
+
+    curr_ref.initRemRegBins = 4;
+    std::generate( curr_ref.numSig, curr_ref.numSig + 4, g6 );
+    std::generate( curr_ref.remRegBins, curr_ref.remRegBins + 4, g11 );
+    for( int idx = 0; idx < 16; ++idx )
+    {
+      std::generate( curr_ref.tplAcc[idx], curr_ref.tplAcc[idx] + 4, g8 );
+      std::generate( curr_ref.sum1st[idx], curr_ref.sum1st[idx] + 4, g8 );
+    }
+    std::fill_n( &curr_ref.absVal[0][0], 16 * 4, 0 ); // AbsVal is always set to 0 prior to updateStates.
```
adamjw24, line 486: You could set `absVal[i][j]` to nonZero for `i > insidePos`, to ensure that they are shuffled properly (and zeroed for `prevId == -1`) during updateState.

test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -253,6 +254,518 @@ class DimensionGenerator
...
+
+    scanInfo.scanIdx = rng.get( 16, 128, 16 );
+    scanInfo.sbbSize = rng.getOneOf<int8_t>( { 4, 16 } );
+    scanInfo.numSbb = 16;
+
+    scanInfo.sbbPos = rng.get( 0, scanInfo.numSbb - 1 );
+    scanInfo.nextSbbRight = 0;
+    scanInfo.nextSbbBelow = 0;
+
+    scanInfo.insidePos = rng.get( 0, 15 );
```
adamjw24, line 591: For `updateStateEOS`, `insidePos` should always be `0`

test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -253,6 +254,518 @@ class DimensionGenerator
...
+    scanInfo.nextInsidePos = rng.get( 0, 15 );
```
adamjw24, line 592: For `updateStateEOS`, `nextInsidePos` should always be `16`

test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -253,6 +254,518 @@ class DimensionGenerator
...
+    scanInfo.gtxCtxOffsetNext = ( int8_t )rng.get( 0, 200 );
+    scanInfo.sigCtxOffsetNext = ( int8_t )rng.get( 0, 200 );
+    scanInfo.currNbInfoSbb.numInv = ( uint8_t )rng.get( 0, 5 );
```
adamjw24, line 595: For `insidePos == 0`, `scanInfo.currNbInfoSbb.numInv` is always equal to `0`

test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -253,6 +254,518 @@ class DimensionGenerator
...
+    std::generate( scanInfo.currNbInfoSbb.invInPos, scanInfo.currNbInfoSbb.invInPos + 5, g4 );
+
+    for( int idx = 0; idx < 4; ++idx )
+    {
+      // prevId >= 4 implies prevId - 4 == idx.
+      decisions.prevId[idx] = rng.getOneOf<int8_t>( { -1, 0, 1, 2, 3, int8_t( idx + 4 ) } );
```
adamjw24, line 601: `updateStateEOS` can also have a `prevId == -2`

adamjw24:
> Thank you for addressing the comments, I think its good to go now.

Sorry, I did look into it a bit more and found minor comments. I guess its nothing critical, so let me know if you want to adapt it or should we merge it for now.

adamjw24 (commented):
.

test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -253,6 +254,518 @@ class DimensionGenerator
...
+  {
+    std::memset( &curr_ref, 0xCD, sizeof( curr_ref ) );
+    std::memset( &curr_opt, 0xCD, sizeof( curr_opt ) );
+
+    scanInfo.insidePos = rng.get( 0, 15 );
+    scanInfo.nextInsidePos = rng.get( 0, 15 );
+    scanInfo.gtxCtxOffsetNext = ( int8_t )rng.get( 0, 200 );
+    scanInfo.sigCtxOffsetNext = ( int8_t )rng.get( 0, 200 );
+    scanInfo.currNbInfoSbb.numInv = ( uint8_t )rng.get( 0, 5 );
+    std::generate( scanInfo.currNbInfoSbb.invInPos, scanInfo.currNbInfoSbb.invInPos + 5, g4 );
```
adamjw24, line 464: Actually, just one last comment, to ensure we don't get undefined behavior. There are some constraints on the variables:
- `insidePos = ( nextInsidePos + 1 ) % sbbSize `
- all `scanInfo.currNbInfoSbb.invInPos` are smaller than `insidePos`
- `scanInfo.currNbInfoSbb.numInv <= std::min(insidePos, 5)`

If the tests function though, I guess its not super important to address this. But maybe the info can also be beneficial for you to develop the NEON SIMD (applying those restrictions will loosen the tests a bit because it would e.g. allow to write values in different order).

athulya-arm: Addressed all comments, please have a look.
Thank you.

test/vvenc_unit_test/vvenc_unit_test.cpp
```diff
@@ -253,6 +254,540 @@ class DimensionGenerator
...
+    scanInfo.sbbSize = rng.getOneOf<int8_t>( { 4, 16 } );
+    scanInfo.nextInsidePos = rng.get( 0, scanInfo.sbbSize - 1 );
+    scanInfo.insidePos = ( scanInfo.nextInsidePos + 1 ) % scanInfo.sbbSize;
+    scanInfo.gtxCtxOffsetNext = ( int8_t )rng.get( 0, 200 );
+    scanInfo.sigCtxOffsetNext = ( int8_t )rng.get( 0, 200 );
+    scanInfo.currNbInfoSbb.numInv = ( uint8_t )rng.get( 0, std::min<int>( scanInfo.insidePos, 5 ) );
+
+    for( int inv = 0; inv < 5; ++inv )
+    {
+      scanInfo.currNbInfoSbb.invInPos[inv] = ( uint8_t )rng.get( 0, std::max<int>( 0, scanInfo.insidePos - 1 ) );
```
adamjw24, line 472: Very strictly, all `invInPos` (inverse template positions within the current subblock) are unique, while the code could generate the same position to be in the list more than once.

I think its a very minor issue, and I will take the code in for now - just keep that in mind for when you're developing the implementation.

adamjw24:
Thank you for bearing with me through this long review process. I think the last minor issue is not super imporant, and I suspect anyway there is more small issues there that I just didn't catch. Please dont hesistate to reach out if you suspect that a problem during the NEON development is caused by the tests themselves and we can discuss here.

Also, please have a look at the following paper, where I documented some of the DQ implementation concepts used here that differ from the reference VTM implementation (like the inverse template).

A. Wieckowski, C. Stoffers, H. Schwarz, B. Bross and D. Marpe, "Fast Dependent Quantization Using Trellis Pruning, Forward Context Adaptation and Vectorization," 2024 32nd European Signal Processing Conference (EUSIPCO), Lyon, France, 2024, pp. 516-520, doi: 10.23919/EUSIPCO63174.2024.10714991.

PR #687, opened 2026-04-08
Arm: Optimize Neon xCalcHAD*

alex-davicenko-arm:
Optimize Neon xCalcHAD* by using `|x+y|+|x-y| = 2*max(|x|,|y|)` identity to fuse last stage of horizontal Hadamard with absolute value summation. Changes to the following kernels provide up to the following uplifts when measured on a Neoverse V2 machine with LLVM-21:
 
- xCalcHAD16x16_fast_Neon: 27%
- xCalcHAD8x8_Neon: 44%
- xCalcHAD8x4_Neon: 16%
- xCalcHAD4x8_neon: 14%
- xCalcHAD4x4_neon: 9%
- xCalcHAD16x8_neon: 43%
- xCalcHAD16x8_neon: 36%

### DIFF
--- a/source/Lib/CommonLib/arm/neon/RdCost_neon.cpp
+++ b/source/Lib/CommonLib/arm/neon/RdCost_neon.cpp
@@ -504,7 +504,7 @@ Distortion xCalcHAD16x8_neon( const Pel* piOrg, const Pel* piCur, int iStrideOrg
   return sad;
 }
 
-static Distortion xCalcHAD8x16_neon( const Pel* piOrg, const Pel* piCur, int iStrideOrg, int iStrideCur )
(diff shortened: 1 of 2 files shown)


adamjw24:
The same trick could be used for the full 16x16 HADs, right? Subsamples HAD is only used in the faster presets.

alex-davicenko-arm:
> The same trick could be used for the full 16x16 HADs, right? Subsamples HAD is only used in the faster presets.

Yes, the Neon implementation does not use a 16x16 HAD kernel but there are patches for 8x8, 8x4, 4x8, 4x4, 16x8 and 8x16 HAD on the way.

adamjw24:
Thank you for updating the improvement for all the functions! Will merge when pipeline has passed

PR #692, opened 2026-04-23
Arm: Add Neon implementation of DepQuant checkAllRdCostsOdd1

micro-arm:
Add a Neon implementation of DepQuant `checkAllRdCostsOdd1`.

This implementation improves overall performance by around 99% compared with the existing SIMDe implementation when benchmarked on a Neoverse V2 using LLVM 22.

### DIFF
--- a/source/Lib/CommonLib/arm/neon/DepQuant_neon.cpp
+++ b/source/Lib/CommonLib/arm/neon/DepQuant_neon.cpp
@@ -332,11 +332,118 @@ void updateStatesEOS_neon( const ScanInfo& scanInfo, const Decisions& decisions,
   updateNextCtx_neon( scanInfo, curr );
 }
 
+static inline uint64x2_t vvenc_vcgtq_s64( int64x2_t a, int64x2_t b )
+{
+#if REAL_TARGET_AARCH64
+  return vcgtq_s64( a, b );
+#else
+  return vreinterpretq_u64_s64( vshrq_n_s64( vqsubq_s64( b, a ), 63 ) );
+#endif
+}
+
+void checkAllRdCostsOdd1_neon( const ScanPosType spt, const int64_t pq_a_dist, const int64_t pq_b_dist,
+                               Decisions& decisions, const StateMem& state )
+{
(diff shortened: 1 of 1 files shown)


adamjw24:
Great! Could you explain a bit how the improvement was achieved? Is there anything that could also be applied to the SSE implementation? Or is it just because the SIMDe translation was so inefficient?

micro-arm:
It's a bit of both: these uplifts usually come from the SIMDe layer being inefficient, so our native implementations are better suited for Arm cores.

Furthermore, I implemented the Neon code to be more faithful to the scalar code, where the outputs are only shuffled at the end. This is possible because the assumptions in the if-else logic at the end of the scalar path always hold true for the initial comparisons:

i.e.
```
// stateId = 0
A[0] < rdCost[2] // always true
Z[0] < rdCost[0] // always true
// stateId = 2
A[2] < rdCost[3] // always true
Z[2] < rdCost[1] // always true

// With these assumptions, we can simplify the output shuffle table as follows:
// Compare A vs Z to set decisions.
rdCost[0]   = ( A[1] < Z[0] ) ? A[1] : Z[0];
absLevel[0] = ( A[1] < Z[0] ) ? 1    : 0;
prevId[0]   = ( A[1] < Z[0] ) ? 1    : 0;

rdCost[1]   = ( A[3] < Z[2] ) ? A[3] : Z[2];
absLevel[1] = ( A[3] < Z[2] ) ? 1    : 0;
prevId[1]   = ( A[3] < Z[2] ) ? 3    : 2;

rdCost[2]   = ( Z[1] < A[0] ) ? Z[1] : A[0];
absLevel[2] = ( Z[1] < A[0] ) ? 0    : 1;
prevId[2]   = ( Z[1] < A[0] ) ? 1    : 0;

rdCost[3]   = ( Z[3] < A[2] ) ? Z[3] : A[2];
absLevel[3] = ( Z[3] < A[2] ) ? 0    : 1;
prevId[3]   = ( Z[3] < A[2] ) ? 3    : 2;
```

adamjw24:
Thanks a lot for the explanation

PR #694, opened 2026-04-28
Arm: fix cap even checks

ylatuya:
Some tests are failing with AArch64 Neon intrinsics enabled after https://github.com/fraunhoferhhi/vvenc/pull/691:

```
[main] Building folder: /Users/andoni/git/vvenc/build
[build] Starting build
[proc] Executing command: /opt/homebrew/bin/cmake --build /Users/andoni/git/vvenc/build --config Debug --target all --
[build] [1/2   0% :: 0.011] Re-checking globbed directories...
[build] -- GLOB mismatch!
[build] [2/2  50% :: 0.027] Re-running CMake...
[build] CMake Warning at /opt/homebrew/Cellar/cmake/3.25.1/share/cmake/Modules/Platform/Darwin-Initialize.cmake:303 (message):
[build]   Ignoring CMAKE_OSX_SYSROOT value:
[build]
[build]    /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.5.sdk
[build]
[build]   because the directory does not exist.
[build] Call Stack (most recent call first):
[build]   /opt/homebrew/Cellar/cmake/3.25.1/share/cmake/Modules/CMakeSystemSpecificInitialize.cmake:21 (include)
[build]   CMakeLists.txt:13 (project)
[build]
[build]
[build] -- CMAKE_MODULE_PATH: updating module path to: /Users/andoni/git/vvenc/cmake/modules
[build] -- normalized target architecture: AARCH64
[build] -- Performing Test SUPPORTED_march=armv8_2_a+sve
[build] -- Performing Test SUPPORTED_march=armv8_2_a+sve - Success
[build] -- Performing Test SUPPORTED_march=armv9_a+sve2
[build] -- Performing Test SUPPORTED_march=armv9_a+sve2 - Success
[build] -- x86 SIMD intrinsics enabled (using SIMDE for non-x86 targets)
[build] -- AArch64 Neon intrinsics enabled
[build] -- AArch64 SVE intrinsics enabled
[build] -- VVENC_ENABLE_THIRDPARTY_JSON: ON
[build] -- Configuring done
[build] -- Generating done
[build] -- Build files have been written to: /Users/andoni/git/vvenc/build
[build] [1/2   0% :: 0.004] Re-checking globbed directories...
[...]

PR #695, opened 2026-04-28
Fix CHECKD condition in Arm DepQuant Neon implementation

georges-arm:
The `CHECKD` conditions are the opposite of normal assert conditions, in particular the expression is when to throw an exception not what is expected to be true.

The `capWithParity_*_neon` functions expect `capEven` to be even but the condition was coded the wrong way around. Swap the condition so that the CHECKD works as expected.


georges-arm:
Duplicate of #694, closing.

Issue #354, opened 2024-02-20
PVS-Studio static code analyzer warnings


birdie-github:
Quite a lot of them, as to whether to fix them or not, it's solely up to you.

[vvenc-PVSStudio-Report.zip](https://github.com/fraunhoferhhi/vvenc/files/14342425/vvenc-PVSStudio-Report.zip)

After opening the report, please first sort by Code then Level.

PR #699, opened 2026-04-29
Fix PVS-Studio V590: simplify redundant isNearVBbelow condition in AdaptiveLoopFilter

kjg0724:
## Summary

PVS-Studio V590 flagged a redundant compound condition in `AdaptiveLoopFilter`:

```cpp
// Before
bool isNearVBbelow = yVb >= vbPos && (yVb <= vbPos);

// After
bool isNearVBbelow = yVb == vbPos;
```

`(yVb >= vbPos) && (yVb <= vbPos)` is logically equivalent to `yVb == vbPos`.
The simplified form is clearer and removes the redundancy.

Fixes all five occurrences:
- `source/Lib/CommonLib/AdaptiveLoopFilter.cpp` (scalar path)
- `source/Lib/CommonLib/x86/AdaptiveLoopFilterX86.h` (4× SIMD paths)

Addresses issue #354.

## Tested

- `cmake --build build --parallel 8` — success (no warnings introduced)

### DIFF
--- a/source/Lib/CommonLib/AdaptiveLoopFilter.cpp
+++ b/source/Lib/CommonLib/AdaptiveLoopFilter.cpp
@@ -893,8 +893,8 @@ void AdaptiveLoopFilter::filterBlk( const AlfClassifier *classifier,
           pImg3 = (yVb <= vbPos + 1) ? pImg1 : pImg3;
           pImg5 = (yVb <= vbPos + 2) ? pImg3 : pImg5;
         }
-        bool isNearVBabove = yVb < vbPos && (yVb >= vbPos - 1);
-        bool isNearVBbelow = yVb >= vbPos && (yVb <= vbPos);
+        bool isNearVBabove = yVb == vbPos - 1;
+        bool isNearVBbelow = yVb == vbPos;
 
         for( int jj = 0; jj < clsSizeX; jj++ )
         {
--- a/source/Lib/CommonLib/x86/AdaptiveLoopFilterX86.h
+++ b/source/Lib/CommonLib/x86/AdaptiveLoopFilterX86.h
@@ -382,8 +382,8 @@ void simdFilter5x5Blk( const AlfClassifier *classifier, const PelUnitBuf& recDst
           pImg1 = (yVb == vbPos) ? pImg0 : pImg1;
           pImg3 = (yVb <= vbPos + 1) ? pImg1 : pImg3;
         }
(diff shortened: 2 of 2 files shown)


K-os:
Did you make sure there is no actual bug hiding behind here? It looks like the code was actually intended to do something else there.

kjg0724:
I had a closer look at this.

`yVb >= vbPos && (yVb <= vbPos)` is mathematically equivalent to
`yVb == vbPos`, so the change is behaviour-identical in all affected
paths.

I also checked the VTM reference (`AdaptiveLoopFilter.cpp`) and the
same condition is there too, unchanged since the initial import — so
this is not a vvenc-specific issue.

Worth noting though: the padding logic further down remaps
`pImg3/pImg4` up to `vbPos + 1` (or `vbPos + 2` for 7x7 luma), while
`isNearVBbelow` only flags `vbPos` for the reduced-precision shift.
That asymmetry was already there before this PR.

Happy to keep this as-is and open a separate issue for the VB condition
if it needs revisiting, or revert if you'd prefer to look at the
underlying logic first.

K-os:
I just looked at it myself, and also saw that the code is exactly the same in VTM, so I assume it is correct. Furthermore, the condition for `isNearVBabove` is also overly complicated and could be simplified to `bool isNearVBabove = (yVb == vbPos - 1);`

I am undecided about how to proceed with this. What do you think @adamjw24 ? Should we keep the code as it is in VTM here, or simplify the conditions? I would assume the compiler does that anyways if it is semantically identical.

adamjw24:
I think it can be simplified.

Now, correct me if I'm wrong, both `isNearVBbelow` and `isNerVBabove` can be simplified:

```cpp
        bool isNearVBabove = yVb < vbPos && (yVb >= vbPos - 1);
        bool isNearVBbelow = yVb >= vbPos && (yVb <= vbPos);
```

to

```cpp
        bool isNearVBabove = yVb == vbPos - 1;
        bool isNearVBbelow = yVb == vbPos;
```

This seems to implement table 45 of the standard text:

grafik

We should check afterwards if the generated bitstreams are bit-equal, but it looks good to me.

kjg0724:
Simplified isNearVBabove the same way — yVb == vbPos - 1, matching the table.

K-os:
You only updated the scalar implementation, not the SIMD implementation.

kjg0724:
Also simplified in the x86 SIMD paths.

PR #701, opened 2026-04-29
Fix PVS-Studio V668: remove redundant post-new null checks

kjg0724:
## Summary

C++ `operator new` throws `std::bad_alloc` on allocation failure; it never
returns `null`.  The null guards added after plain `new` expressions are
unreachable dead code and are flagged by PVS-Studio V668.

**vvenc.cpp** — `vvenc_encoder_create()`:
```cpp
// Before
vvenc::VVEncImpl* encCtx = new vvenc::VVEncImpl();
if (!encCtx) { return nullptr; }
return (vvencEncoder*)encCtx;

// After
vvenc::VVEncImpl* encCtx = new vvenc::VVEncImpl();
return (vvencEncoder*)encCtx;
```

**TrQuant.cpp** — `TrQuant::setNewQPBlaMode()`:
```cpp
// Before
{ m_quant = new DepQuant(otherQuant, bEnc, scalingListsEnabled); }
if (m_quant) { m_quant->init(rdoq, bUseRDOQTS, thrVal); }

// After
m_quant = new DepQuant(otherQuant, bEnc, scalingListsEnabled);
m_quant->init(rdoq, bUseRDOQTS, thrVal);
```

No behaviour change.

Addresses issue #354.

## Tested

- `cmake --build build --parallel 8` — success

### DIFF
--- a/source/Lib/CommonLib/TrQuant.cpp
+++ b/source/Lib/CommonLib/TrQuant.cpp
@@ -309,14 +309,9 @@ void TrQuant::init( const Quant* otherQuant,
   delete m_quant;
   m_quant = nullptr;
 
-  {
-    m_quant = new DepQuant( otherQuant, bEnc, scalingListsEnabled );
-  }
-
-  if( m_quant )
-  {
-    m_quant->init( rdoq, bUseRDOQTS, thrVal );
-  }
+  m_quant = new(std::nothrow) DepQuant( otherQuant, bEnc, scalingListsEnabled );
+  CHECK( !m_quant, "allocation failed" );
+  m_quant->init( rdoq, bUseRDOQTS, thrVal );
 }
 
 
--- a/source/Lib/vvenc/vvenc.cpp
+++ b/source/Lib/vvenc/vvenc.cpp
@@ -214,13 +214,7 @@ VVENC_DECL void vvenc_accessUnit_default(vvencAccessUnit *accessUnit )
 
 VVENC_DECL vvencEncoder* vvenc_encoder_create()
 {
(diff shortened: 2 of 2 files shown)


K-os:
Thanks for taking the time to look into the PVS-Studio warnings, but I think there is no value in blindly fixing warnings, without correcting the underlying issue.

Yes, the existing code here was incorrect, but just eliminating the faulty error handling code has no value in itself. If you touch that code, please fix it correctly. So here the correct fix would be to either actually catch the exception or to use `new(std::nothrow)` for allocation.

K-os (changes requested):
Ok, in vvenc.cpp the change is correct, because it prevents the C++ exception from,  leaving the C++ code and being thrown out to consumers of the library. But just bypassing `DepQuant::init()` in TrQuant.cpp is probably not what we want.

A CHECK( !m_quant, "allocation failed" ) or something like that would probably be better, because that exception should be caught within VVenC.

kjg0724:
Good point. Updated:

- `vvenc_encoder_create`: switched to `new(std::nothrow)` so an
  allocation failure returns `nullptr` to the C caller.
  `vvenc_encoder_open` already handles a null pointer.

- `TrQuant::init`: same change — `new(std::nothrow)` makes the
  null-guard meaningful instead of dead code.

Used `nothrow` in both places to stay consistent with the
`#if HANDLE_EXCEPTION` convention in `vvencimpl.cpp`. Build and
tests still pass.

kjg0724:
Switched to new(std::nothrow) + CHECK as suggested — throws on allocation failure instead of silently skipping init.

K-os (approved):

PR #702, opened 2026-04-29
Fix PVS-Studio V663/V1024: replace while(!eof) with getline loop in IStreamIO.h

kjg0724:
## Summary

`while (!in.eof())` followed by `std::getline()` is a well-known C++ anti-pattern
(PVS-Studio V663 / V1024): `eofbit` is only set *after* a failed read, so the
loop can process a spurious empty line after the last real token, or miss
detection of read errors entirely.

The idiomatic fix is to let `getline` control the loop condition directly:

```cpp
// Before
while ( ! in.eof() )
{
    std::string line;
    std::getline( in, line );
    ...
}

// After
for ( std::string line; std::getline( in, line ); )
{
    ...
}
```

Four occurrences fixed (all in `source/Lib/apputils/IStreamIO.h`):

| Line | Operator |
|------|----------|
| 107 | `operator>>` for `IStreamToRefVec<T>` |
| 389 | `operator>>` for `IStreamToVec<T>` |
| 501 | `operator>>` for `IStreamToArr<T>` |
| 554 | `operator>>` for `IStreamToArr<char>` |

Behaviour is preserved for all normal inputs.  The change eliminates the
risk of a spurious empty-string iteration when the stream ends with a newline.

Addresses issue #354.

## Tested

- `cmake --build build --parallel 8` — success

### DIFF
--- a/source/Lib/apputils/IStreamIO.h
+++ b/source/Lib/apputils/IStreamIO.h
@@ -104,10 +104,8 @@ inline std::istream& operator >> ( std::istream& in, IStreamToRefVec<T>& toVec )
   size_t idx = 0;
   bool fail = false;
   // split into multiple lines if any
-  while ( ! in.eof() )
+  for ( std::string line; std::getline( in, line ); )
   {
-    std::string line;
-    std::getline( in, line );
     // treat all whitespaces and commas as valid separators
     if( toVec.sep == 'x')
       replace_if( line.begin(), line.end(), []( int c ){ return isspace( c ) || c == 'x'; }, ' ' );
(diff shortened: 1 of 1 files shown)


K-os:
I'm sorry I don't like this solution. You fixed a non-existent problem with a lot of overly complicated code to circumvent an error, that you added in your first commit.
There was no spurious empty line read and even if it was, there would have been no problem processing an empty line in the loop.

I acknowledge that the whole code is not very pretty, but throwing more bad code at it does not help.

kjg0724:
Right — the empty-line guard inside the loop makes the extra iteration harmless.

PR #705, opened 2026-05-05
Add templated implementation for gnsSolveByChol

athulya-arm:
Add templated Neon implementations for ALF orders 7 and 13 for
AlfCovariance::gnsSolveByChol.
AlfCovariance::gnsSolveByChol is only reached with orders 7 or 13 in
encoder ALF paths - 7 coefficients for chroma/CC-ALF-style filters and
13 coefficients for luma filters.

Move the templated Cholesky helper routines out of class AlfCovariance
and keep them as static internal helpers in EncAdaptiveLoopFilter.cpp.

The templated path may produce small floating-point differences from
the existing version because compiler optimizations can change the
order of operations, causing tiny numeric variation even when the
algorithm is the same.

### DIFF
--- a/source/Lib/EncoderLib/EncAdaptiveLoopFilter.cpp
+++ b/source/Lib/EncoderLib/EncAdaptiveLoopFilter.cpp
@@ -883,13 +883,16 @@ alf_float_t AlfCovariance::calculateError( const int *clip ) const
 #define REG_SQR          alf_float_t( 0.0000001 )
 
 //Find filter coeff related
-int AlfCovariance::gnsCholeskyDec( TE inpMatr, TE outMatr, int numEq ) const
+template<int NumEq>
+static int gnsCholeskyDec( AlfCovariance::TE inpMatr, AlfCovariance::TE outMatr )
 {
-  for( int i = 0; i < numEq; i++ )
+  static_assert( NumEq == 7 || NumEq == 13, "Unsupported ALF Cholesky order" );
+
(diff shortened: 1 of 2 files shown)


K-os:
I don't really understand it: I don't see any NEON-specific code, but you say it performs 15% better than the original implementation. Is it because the compiler can better optimize the templated implementations? Or am I missing something?

If it's due to the compiler optimizations, I would assume this applies to other architectures as well. Wouldn't it make sense then, to use the templated implementation for all architectures?

athulya-arm:
As I mentioned in the commit message, this patch is an incremental step toward a full Neon implementation, the complete Neon specific version will come in as a follow-up patch. The current improvement comes from better compiler optimizations when the template order is known at compile time. 

This approach could benefit other architectures as well - for x86, a small improvement is observed in both CPU time and elapsed time. We can update the patch accordingly to make the templated implementations generic for all architectures.

K-os:
Thank you very much for the update.
So the fallback using the non-templated implementation is not needed any more? I think then it really makes sense to use the template implementation for all architectures.

athulya-arm:
Yes, the function is only reached with orders 7 or 13 in encoder ALF paths - 7 coefficients for chroma/CC-ALF-style filters and 13 coefficients for luma filters. So the non templated fallback is not necessary.

Since the patch is already merged, it would still be good to hear whether you observed similar gains on other architectures or if you have any concerns about removing that fallback entirely.

PR #707, opened 2026-05-13
Arm: Add Neon implementation for gnsSolveByChol

athulya-arm:
- Add Neon implementation for `AlfCovariance::gnsSolveByChol`, with the optimization work mainly focused on ALF order 13. The output `md5sum` values may differ from the existing implementation because of floating-point accuracy differences.

Add Arm initialization and function-pointer dispatch for `AlfCovariance::gnsSolveByChol`. Update CMakeLists to include Arm Neon sources from new EncoderLib location.
Make `gnsSolveByChol` static so they can be used through plain function-pointer dispatch and called without an `AlfCovariance` instance; the previous const qualifier on the function are no longer needed as it is static member function now.

This new implementation improves performance by approximately ~20% for order 13 compared to the generic version when benchmarked on a Neoverse V2 with LLVM 21.

- Add unit tests for `AlfCovariance::gnsSolveByChol`.

Add `FloatGenerator` for floating point inputs and tolerance in `compare_values_1d` for comparing results.
Modify the AlfCovariance constructor to include a new `enableOpt` parameter to create a reference object and an optimized object for testing.

### DIFF
--- a/source/Lib/CommonLib/arm/InitARM.cpp
+++ b/source/Lib/CommonLib/arm/InitARM.cpp
@@ -60,6 +60,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "CommonLib/AffineGradientSearch.h"
 #include "CommonLib/AdaptiveLoopFilter.h"
 #include "CommonLib/SampleAdaptiveOffset.h"
+#include "EncoderLib/EncAdaptiveLoopFilter.h"
 
 namespace vvenc
 {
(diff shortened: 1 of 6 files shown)


adamjw24:
I see you added a few algorithmic tricks. How often is the `checkForZero`-check used? Would it be worth to include it in the X86 implementation? Also, looks good to me. I guess as a next step we could do the same to `calculateErrorForCoeffs` and remove the inline SIMD from this file.

athulya-arm:
In gnsSolveByChol, there is a chance for the input matrix to be all zero. So I added a separate path to check for this case, clear the output, and return early.
I collected statistics for two streams, measuring the AllZero hits as a percentage of the total calls to gnsSolveByChol

Stream1: Bosphorus_3840x2160.y4m
  Order 7 : 50.90%
  Order 13: 23.44%

Stream2: Bosphorus_1920x1080_120fps_420_8bit_YUV.y4m
  Order 7 : 33.99%
  Order 13: 0.64%

Based on this, the check for zero seems useful, especially for order 7. I believe it could be considered for the x86 implementation as well, though it will add a few extra cycles for Cholesky fails or where Cholesky with delta cases.

PR #710, opened 2026-06-17
Refactor EncAdaptiveLoopFilter::getPreBlkStats

athulya-arm:
Refactor EncAdaptiveLoopFilter::getPreBlkStats

    Refactor the weighted and unweighted covariance accumulation paths out
    of `EncAdaptiveLoopFilter::getPreBlkStats()` into dispatchable helper
    functions.
    
    Introduce function pointers `m_getPreBlkStatsAccum` and
    `m_getPreBlkStatsWeightedAccum` to `EncAdaptiveLoopFilter` class.
    Move the x86 implementations into the EncoderLib x86 SIMD layer and
    assign them during EncAdaptiveLoopFilter x86 initialization, while
    keeping scalar helpers as the default fallback.
    
    Also include the EncoderLib x86 sources in the VVenC CMake x86 file
    lists.

Add unit tests for EncAdaptiveLoopFilter::getPreBlkStats functions.
    
    Modify the EncAdaptiveLoopFilter constructor to include a new
    'enableOpt' parameter to create a reference object and an optimized
    object for testing.
    
    Add optional `tolerance` handling to `compare_value` so unit tests can
    compare results that are numerically equivalent but not bit-exact. This
    allows small floating-point accuracy differences in
    `getPreBlkStatsWeightedAccum` test cases. SIMD implementations can
    change the order of floating-point operations compared with the scalar
    reference, which may produce tiny differences while still being
    mathematically correct.

### DIFF
--- a/source/Lib/CommonLib/x86/InitX86.cpp
+++ b/source/Lib/CommonLib/x86/InitX86.cpp
@@ -62,6 +62,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "TrQuant_EMT.h"
 #include "QuantRDOQ2.h"
 #include "SEIFilmGrainAnalyzer.h"
+#include "EncoderLib/EncAdaptiveLoopFilter.h"
 
 #ifdef TARGET_SIMD_X86
 
@@ -205,6 +206,22 @@ void AdaptiveLoopFilter::initAdaptiveLoopFilterX86()
     break;
   default:
     break;
+  }
+}
+
+void EncAdaptiveLoopFilter::initEncAdaptiveLoopFilter_X86()
+{
+  auto vext = read_x86_extension_flags();
+  switch (vext){
+    case AVX512:
+    case AVX2:
+    case AVX:
+    case SSE42:
+    case SSE41:
+      _initEncAdaptiveLoopFilter_X86<SSE41>();
+      break;
+    default:
+      break;
   }
 }
 #endif
--- a/source/Lib/EncoderLib/EncAdaptiveLoopFilter.cpp
+++ b/source/Lib/EncoderLib/EncAdaptiveLoopFilter.cpp
@@ -1052,6 +1052,11 @@ int AlfCovariance::gnsSolveByChol( TE LHS, alf_float_t* rhs, alf_float_t*x, int 
 }
 //////////////////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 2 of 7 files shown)


athulya-arm:
This is an initial commit that extracts `EncAdaptiveLoopFilter::getPreBlkStats` as a standalone function. A Neon implementation of the same will be added in a subsequent patch.

K-os:
Hi, thank you very much for the patch. In principle, the introduced structure looks good to me, but I am not sure if there was a specific reason why the SIMD code was implemented inline here, or if maybe the added indirect function call adds too much overhead here. We'll talk about that internally and I'll get back to you.

athulya-arm:
Hi, just a gentle follow-up on this PR. I was wondering if there have been any updates from your internal discussion. Whenever you have a chance, I would appreciate any feedback. Thank you.

adamjw24:
Looks good to me. I was just being lazy back in the day.

Could you add the following to the patch:

```patch
diff --git a/source/Lib/vvenc/CMakeLists.txt b/source/Lib/vvenc/CMakeLists.txt
index bd29c727..0ab179da 100644
--- a/source/Lib/vvenc/CMakeLists.txt
+++ b/source/Lib/vvenc/CMakeLists.txt
@@ -119,7 +119,6 @@ if( VVENC_ENABLE_X86_SIMD )
     set_property( SOURCE ${AVX_SRC_FILES}   APPEND PROPERTY COMPILE_FLAGS "${FLAG_mavx}"   )
     set_property( SOURCE ${AVX2_SRC_FILES}  APPEND PROPERTY COMPILE_FLAGS "${FLAG_mavx2}"  )

-    set_source_files_properties( ../EncoderLib/EncAdaptiveLoopFilter.cpp APPEND PROPERTY COMPILE_FLAGS "${FLAG_msse41}" )
     set_source_files_properties( ../CommonLib/LoopFilter.cpp APPEND PROPERTY COMPILE_FLAGS "${FLAG_msse41}" )
     set_source_files_properties( ../CommonLib/QuantRDOQ2.cpp APPEND PROPERTY COMPILE_FLAGS "${FLAG_msse41}" )
   endif()
```

adamjw24:
Wait, no, there's still `calcErrorForCoeffsLin_13_SSE`, so I guess its ok for now.

Issue #711, opened 2026-06-22
arm/neon: SIMD coverage gap — Quant and FGA have no Neon implementation


kjg0724:
## Intent

Following my previous ARM NEON work (xFilteringPandQ in #703, #698), I went looking for the next gap and found **two CommonLib modules that have x86 SIMD (SSE4.1/AVX2) but no ARM NEON**, so they fall back to scalar on AArch64:

| Module | Function pointers | x86 reference |
|---|---|---|
| **Quant** | `xQuant` / `xDeQuant` / `xNeedRdoq` (`Quant.h`) | `x86/QuantX86.h` (~600 lines) |
| **FGA** (Film Grain Analysis) | Canny `gradient`, morphology `dilation`/`erosion`, analyzer (`SEIFilmGrainAnalyzer.h`) | `x86/FGAX86.h` (~1000 lines) |

Neither has a NEON counterpart in VVdeC, so these would be native ports mirroring the existing x86 path, following the established ARM porting pattern (RdCost, DepQuant, MCTF, etc.).

## Question (to avoid duplicate effort)

Posting here as with #698 — are either of these being worked on internally at Fraunhofer or by the Arm team?

If free, I'd like to start with **FGA** (it looks furthest from the Arm team's current quant/RDOQ work), but I'm happy to take **Quant** instead if FGA is spoken for. Please let me know what's available.

As before, the PR will include: full path coverage (not a partial scalar fallback), a ref-vs-opt unit test in `vvenc_unit_test.cpp`, and benchmark numbers.

cc @adamjw24 @georges-arm @athulya-arm @micro-arm @alex-davicenko-arm @lizhang-arm

georges-arm:
> Posting here as with [#698](https://github.com/fraunhoferhhi/vvenc/issues/698) — are either of these being worked on internally at Fraunhofer or by the Arm team?

Hi! Thanks for your help in porting some of the functions to AArch64. I can confirm nobody from the Arm side is looking at the ones you mentioned (Quant/FGA).

kjg0724:
Thanks @georges-arm, that's great to hear. I'll start with **FGA** then, beginning with the smaller `FGAnalyzer::calcVar`/`calcMean` helpers and working up to `Morph::dilation` and `Canny::gradient`, each as a focused PR with a unit test and benchmarks. Will keep Quant as a follow-up.

kjg0724:
[...]

PR #703, opened 2026-04-29
arm/neon: add Neon long deblocking filter

kjg0724:
Adds an ARM Neon path for the long (bilinear) deblocking filter
(xFilteringPandQ), registered via initLoopFilterARM() alongside the
existing x86 path.

Horizontal edges (step == 1):
Uses contiguous vld1_s16 loads and processes four rows in parallel.
Reference values are prepared per row, then the weighted blend and
tc-clip are applied with Neon.

Vertical edges (step == stride):
Uses the same vmull/vmlal int16x4_t arithmetic as the horizontal path.
Lane loads/stores (vld1_lane_s16 / vst1_lane_s16) handle the
non-contiguous vertical access pattern.

Performance (Neoverse N1, clang 15, N=2,000,000, 7-round median):
hor 7+7: scalar 206.3ns -> neon 90.6ns, 2.28x
hor 5+5: 163.6 -> 68.5, 2.39x
hor 7+5: 188.9 -> 82.6, 2.29x
hor 7+3: 170.0 -> 75.3, 2.26x
hor 3+5: 133.6 -> 60.2, 2.22x
ver 7+7: 206.2 -> 130.3, 1.58x
ver 5+5: 161.6 -> 98.4, 1.64x
ver 7+5: 182.6 -> 117.4, 1.56x
ver 7+3: 171.9 -> 105.7, 1.63x
ver 3+5: 138.0 -> 84.5, 1.63x

Tested:
- aarch64 build
- ctest
- LoopFilterPandQ unit test in vvenc_unit_test.cpp, covering all nP/nQ
  combinations for horizontal and vertical filters with deterministic
  and random inputs

PR #713, opened 2026-06-24
Arm: Add Neon implementation for FGA calcVar and calcMean

kjg0724:
Add Neon implementations for the FGAnalyzer calcVar/calcMean hooks,
which are called during FGA parameter estimation on analyzed frames
when `--fga` is enabled.

## Implementation

`calcMeanNeon` accumulates pixel sums directly into `int32x4_t` via
`vpaddlq_s16` per load, using a 16-pixel unrolled main loop with two
independent accumulators to expose instruction-level parallelism. The
outer loop advances the row pointer by `origStride` instead of
multiplying per iteration.

`calcVarNeon` follows the same two-pass structure as `calcVarSse`
(mean first, then variance). The mean pass uses the same dual-
accumulator pattern. The variance pass uses two independent `int64x2_t`
accumulators to reduce add-chain latency, matching the
`_mm_madd_epi16` + `_mm_cvtepi32_epi64` semantics of the SSE path.

## Performance

Benchmarked on Neoverse N1 (GCP T2A, 2 vCPU, Ubuntu 22.04,
gcc -O3 -mcpu=native), median of 7 runs, single core:

| Block size | calcVar scalar→Neon | calcMean scalar→Neon |
|------------|--------------------|--------------------|
| 8×8        | 49.0 → 41.5 ns (1.18×) | 18.4 → 17.5 ns (1.05×) |
| 16×16      | 135.3 → 116.0 ns (1.17×) | 43.0 → 32.6 ns (1.32×) |
| 32×32      | 451.9 → 422.3 ns (1.07×) | 136.8 → 114.1 ns (1.20×) |
| 64×64      | 1678.3 → 1511.7 ns (1.11×) | 395.5 → 309.9 ns (1.28×) |

## Testing

Unit test `FGAnalyzer` added to `vvenc_unit_test`, covering all
combinations of width and height in {8, 16, 32, 64} with 100 random
10-bit inputs (10 in fast mode), verified bit-exact against the scalar
reference (tolerance = 0). Tested on Apple M1 and Neoverse N1.

Relates to #711.

### DIFF
--- a/source/Lib/CommonLib/arm/neon/FGA_neon.cpp
+++ b/source/Lib/CommonLib/arm/neon/FGA_neon.cpp
@@ -133,17 +133,6 @@ void FGAnalyzer::_initFGAnalyzerARM()
 
 template void FGAnalyzer::_initFGAnalyzerARM<NEON>();
 
-// Entry points for the unit test (the function pointers are private to init).
-double calcVarNeonEntry( const Pel* org, const ptrdiff_t origStride, const int w, const int h )
-{
-  return calcVarNeon( org, origStride, w, h );
-}
-
-int calcMeanNeonEntry( const Pel* org, const ptrdiff_t origStride, const int w, const int h )
-{
-  return calcMeanNeon( org, origStride, w, h );
-}
-
 }  // namespace vvenc
 
 #endif  // TARGET_SIMD_ARM && ENABLE_SIMD_OPT_FGA
--- a/source/Lib/CommonLib/arm/neon/FGA_neon.cpp
+++ b/source/Lib/CommonLib/arm/neon/FGA_neon.cpp
@@ -1,0 +1,149 @@
+/* -----------------------------------------------------------------------------
+The copyright in this software is being made available under the Clear BSD
+License, included below. No patent rights, trademark rights and/or
+other Intellectual Property Rights other than the copyrights concerning
+the Software are granted under this license.
+
+The Clear BSD License
+
+Copyright (c) 2019-2026, Fraunhofer-Gesellschaft zur Förderung der angewandten Forschung e.V. & The VVenC Authors.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification,
+are permitted (subject to the limitations in the disclaimer below) provided that
+the following conditions are met:
+
+     * Redistributions of source code must retain the above copyright notice,
+     this list of conditions and the following disclaimer.
+
+     * Redistributions in binary form must reproduce the above copyright
+     notice, this list of conditions and the following disclaimer in the
+     documentation and/or other materials provided with the distribution.
+
+     * Neither the name of the copyright holder nor the names of its
+     contributors may be used to endorse or promote products derived from this
+     software without specific prior written permission.
+
+NO EXPRESS OR IMPLIED LICENSES TO ANY PARTY'S PATENT RIGHTS ARE GRANTED BY
+THIS LICENSE. THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND
+CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
+CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+------------------------------------------------------------------------------------------- */
+
+/** \file     FGA_neon.cpp
+    \brief    film grain analysis, Neon version
+*/
+
+#include "CommonDefARM.h"
+#include "CommonLib/CommonDef.h"
+#include "CommonLib/Rom.h"
+#include "EncoderLib/SEIFilmGrainAnalyzer.h"
+#include "sum_neon.h"
+
+#include <arm_neon.h>
+
+#if defined( TARGET_SIMD_ARM ) && ENABLE_SIMD_OPT_FGA
+
+namespace vvenc
+{
+
+// Mirrors calcMeanSse: returns the integer sum of all w*h pixels (w is a
+// multiple of 8). The caller divides by the block size.
+static int calcMeanNeon( const Pel* org, const ptrdiff_t origStride, const int w, const int h )
+{
+  // Two independent accumulators allow out-of-order execution to overlap
+  // the add chains and avoid the per-row int16 accumulator reset overhead.
+  int32x4_t acc0 = vdupq_n_s32( 0 );
+  int32x4_t acc1 = vdupq_n_s32( 0 );
+  for( int y = 0; y < h; y++, org += origStride )
+  {
+    int x = 0;
+    for( ; x < w - 8; x += 16 )
+    {
+      acc0 = vaddq_s32( acc0, vpaddlq_s16( vld1q_s16( org + x ) ) );
+      acc1 = vaddq_s32( acc1, vpaddlq_s16( vld1q_s16( org + x + 8 ) ) );
+    }
+    for( ; x < w; x += 8 )
+      acc0 = vaddq_s32( acc0, vpaddlq_s16( vld1q_s16( org + x ) ) );
+  }
+  return horizontal_add_s32x4( vaddq_s32( acc0, acc1 ) );
+}
+
+// Mirrors calcVarSse: integer accumulation throughout, single final divide.
+static double calcVarNeon( const Pel* org, const ptrdiff_t origStride, const int w, const int h )
+{
+  // Pass 1: mean — same dual-accumulator pattern as calcMean.
+  int32x4_t avgAcc0 = vdupq_n_s32( 0 );
+  int32x4_t avgAcc1 = vdupq_n_s32( 0 );
+  const Pel* p      = org;
+  for( int y = 0; y < h; y++, p += origStride )
+  {
+    int x = 0;
+    for( ; x < w - 8; x += 16 )
+    {
+      avgAcc0 = vaddq_s32( avgAcc0, vpaddlq_s16( vld1q_s16( p + x ) ) );
+      avgAcc1 = vaddq_s32( avgAcc1, vpaddlq_s16( vld1q_s16( p + x + 8 ) ) );
+    }
+    for( ; x < w; x += 8 )
+      avgAcc0 = vaddq_s32( avgAcc0, vpaddlq_s16( vld1q_s16( p + x ) ) );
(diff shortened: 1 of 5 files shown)


source/Lib/CommonLib/arm/neon/FGA_neon.cpp
```diff
@@ -0,0 +1,149 @@
...
+  // Two independent accumulators allow out-of-order execution to overlap
+  // the add chains and avoid the per-row int16 accumulator reset overhead.
+  int32x4_t acc0 = vdupq_n_s32( 0 );
+  int32x4_t acc1 = vdupq_n_s32( 0 );
+  for( int y = 0; y < h; y++, org += origStride )
+  {
+    int x = 0;
+    for( ; x < w - 8; x += 16 )
+    {
+      acc0 = vaddq_s32( acc0, vpaddlq_s16( vld1q_s16( org + x ) ) );
```
georges-arm: You could try using `vpadalq_s16` here to perform the accumulation in the same step as the pairwise addition. Performance of this instruction is suboptimal up-to-and-including Neoverse V2 but should be improved from Neoverse V3 and later.

If the input data does not fill the `int16` input you might also be able to do some normal `vaddq_s16` accumulation steps before needing to widen. This might not change performance though especially on later CPUs.

Same for the `calcVarNeon` further down.

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
```diff
@@ -0,0 +1,149 @@
...
+      avgAcc0 = vaddq_s32( avgAcc0, vpaddlq_s16( vld1q_s16( p + x ) ) );
+      avgAcc1 = vaddq_s32( avgAcc1, vpaddlq_s16( vld1q_s16( p + x + 8 ) ) );
+    }
+    for( ; x < w; x += 8 )
+      avgAcc0 = vaddq_s32( avgAcc0, vpaddlq_s16( vld1q_s16( p + x ) ) );
+  }
+  const int shift    = Log2( w ) + Log2( h ) - 4;
+  const int avg      = horizontal_add_s32x4( vaddq_s32( avgAcc0, avgAcc1 ) ) >> shift;
+  // Match _mm_packs_epi32 saturation to the int16 range.
+  const int16x8_t vavg = vdupq_n_s16( (int16_t)Clip3( -32768, 32767, avg ) );
```
georges-arm, line 104: You can use `vqmovns_s32` to do the clipping I think.

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
```diff
@@ -0,0 +1,149 @@
...
+  {
+    for( int x = 0; x < w; x += 8 )
+    {
+      int16x8_t pix = vshlq_n_s16( vld1q_s16( p + x ), 4 );
+      pix           = vsubq_s16( pix, vavg );
+      const int32x4_t lo = vmull_s16( vget_low_s16( pix ), vget_low_s16( pix ) );
+      const int32x4_t hi = vmull_s16( vget_high_s16( pix ), vget_high_s16( pix ) );
+      // Match _mm_madd_epi16 + _mm_cvtepi32_epi64: add adjacent products in
+      // 32-bit (wrapping), then sign-extend to 64-bit before accumulating.
+      var0 = vaddq_s64( var0, vmovl_s32( vpadd_s32( vget_low_s32( lo ), vget_high_s32( lo ) ) ) );
```
georges-arm: You can use `vaddw_s32` to avoid the separate `movl_s32` step, or just `vpadalq_s32` to do the pairwise sum and accumulate in a single step (you may want to unroll the loop in this case to be able to use a second accumulator still).

source/Lib/EncoderLib/SEIFilmGrainAnalyzer.cpp
```diff
@@ -752,6 +752,8 @@ void FGAnalyzer::init ( const int width,
 
 #if ENABLE_SIMD_OPT_FGA && defined( TARGET_SIMD_X86 )
   initFGAnalyzerX86();
+#elif ENABLE_SIMD_OPT_FGA && defined( TARGET_SIMD_ARM )
+  initFGAnalyzerARM();
```
georges-arm: I think this should be something like:

```
#if ENABLE_SIMD_OPT_FGA && defined( TARGET_SIMD_X86 )
  initFGAnalyzerX86();
#endif
#if ENABLE_SIMD_OPT_FGA && defined( TARGET_SIMD_ARM )
  initFGAnalyzerARM();
#endif
```

So that we are able to call both `init*` functions to enable us to continue to use any functions that have a SIMDe implementation and no Neon implementation, while preferring the Neon implementations where present.

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
```diff
@@ -0,0 +1,149 @@
...
+  int64x2_t var1 = vdupq_n_s64( 0 );
+  p = org;
+  for( int y = 0; y < h; y++, p += origStride )
+  {
+    for( int x = 0; x < w; x += 8 )
+    {
+      int16x8_t pix = vshlq_n_s16( vld1q_s16( p + x ), 4 );
+      pix           = vsubq_s16( pix, vavg );
+      const int32x4_t lo = vmull_s16( vget_low_s16( pix ), vget_low_s16( pix ) );
+      const int32x4_t hi = vmull_s16( vget_high_s16( pix ), vget_high_s16( pix ) );
```
georges-arm, line 118: Not this PR, but you may want to try the SVE dot-product instructions if you have access to a machine that supports it, see: https://github.com/fraunhoferhhi/vvenc/blob/master/source/Lib/CommonLib/arm/sve/neon_sve_bridge.h#L64

The dot product instructions I linked could be used to multiply and accumulate groups of four pairs of entries. It widens to `int64x2_t` natively so would save the widening step as well.

georges-arm (commented):
Thanks for the Arm code!

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
kjg0724: Good suggestion — applied `vpadalq_s16` in both `calcMeanNeon` and the mean pass of `calcVarNeon`. The per-row `vaddq_s16` accumulation before widening would also work (each lane accumulates at most 8 × 1023 = 8184, well within `int16`), but the single-load-per-iteration pattern is retained for simplicity.

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
kjg0724: `vqmovns_s32` is AArch64-only and the project also builds for armv7 (`ubuntu-24.04: arm-linux-gnueabihf-g++` in CI), so `Clip3` is kept for portability.

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
kjg0724: Applied `vpadalq_s32` (SADALP). It replaces the `vpadd_s32` + `vmovl_s32` + `vaddq_s64` sequence with a single pairwise long accumulate: each `int32` element is sign-extended to `int64` before the adjacent-pair addition, so there is no 32-bit overflow concern.

Updated benchmark on Neoverse N1 (GCP T2A, gcc -O3 -mcpu=native, median of 7 runs):

| Block size | calcVar scalar→Neon | calcMean scalar→Neon |
|------------|--------------------|--------------------|
| 8×8        | 49.0 → 34.0 ns (1.44×) | 18.4 → 17.2 ns (1.07×) |
| 16×16      | 135.3 → 87.8 ns (1.54×) | 42.8 → 29.3 ns (1.46×) |
| 32×32      | 451.9 → 316.3 ns (1.43×) | 136.9 → 113.7 ns (1.20×) |
| 64×64      | 1675.5 → 1071.3 ns (1.56×) | 395.0 → 307.6 ns (1.28×) |

source/Lib/EncoderLib/SEIFilmGrainAnalyzer.cpp
kjg0724: Fixed — split into independent `#if` blocks so that future FGA functions with a SIMDe implementation but no Neon counterpart will not be disabled on ARM.

K-os:
@georges-arm thank you very much for your thorough review. Can you please give your approval, when you think the PR is good to go?

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
georges-arm: Good point, ok!

georges-arm (commented):
I can't comment on the validity of the test cases but the Neon intrinsics code LGTM now. Thanks for the changes!

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
kjg0724: Thanks for the tip, noted for a follow up SVE PR.

K-os:
> I can't comment on the validity of the test cases but the Neon intrinsics code LGTM now. Thanks for the changes!

Yes, that was my intent. Thank you.

I just realized the x86 film grain implementation uses the `calcVarSse` implementation from the MCTF code. Wouldn't it make sense to move the calcVarNeon implementation there and use it for both also?

Furthermore, defining the functions in SEIFilmGrainAnalyzer.cpp outside of the `namespace vvenc` was unintentional. Can you please correct this omission, and adapt the declarations ins vvenc_unit_test accordingly?

K-os:
Actually, thinking about it: why are the functions declared in vvenc_unit_test.cpp at all? Can't we just use the same convention as all other unit tests and initialize one instance of the FGAnalyzer with, and one without optimizations?

The only other unit test that departs from that convention is `test_LoopFilterPandQ` written by you (#703), which I didn't realize at the time of merging.

kjg0724:
Done. Moved `calcVar`, `calcMean`, and `initFGAnalyzerARM()` to public and initialized the scalar pointers in the constructor, so the test can just create two `FGAnalyzer` instances the same way `PelBufferOps` does. Shims are gone.

K-os:
Sorry, PelBufferOps is not the best example. Look at the implementation of IntraPrediction for example: the classes with ARM SIMD implementations built in recent months take a parameter `enableOpt` and then the constructor initializes the function pointers accordingly.
The test cases don't need to know whether the optimized implementation under test is the ARM or x86 version.

What about using the `calcVarNeon` implementation for both MCTF and FGA?

kjg0724:
Makes sense. I'll add `bool enableOpt = true` to the constructor and move the SIMD init there, same as MCTF. The test can then drop the `TARGET_SIMD_ARM` guard and just use `FGAnalyzer{false}` / `FGAnalyzer{true}`.

On sharing `calcVarNeon` between MCTF and FGA: the implementations are structurally identical, so I agree that's a good idea. I'd prefer to handle it in a follow-up PR to keep this one focused - is that ok?

K-os:
Ok, do it as you like.

source/Lib/EncoderLib/SEIFilmGrainAnalyzer.cpp
```diff
@@ -746,14 +758,8 @@ void FGAnalyzer::init ( const int width,
     m_DCTtemp = ( TCoeff* ) xMalloc( TCoeff, DATA_BASE_SIZE * DATA_BASE_SIZE );
   }
 
-  calcVar=calcVarCore;
-  calcMean=calcMeanCore;
   fastDCT2_64 = fastForwardDCT2_B64;
```
K-os: Can you please move this line to the constructor also? It overrides whatever was set there through initFGAnalyzerX86(). Furthermore, from a code structure perspective it also makes sense to have these function pointers all initialized in the same place.

source/Lib/EncoderLib/SEIFilmGrainAnalyzer.h
K-os: I think for consistency this should also be moved to the public section.

K-os (changes requested):

source/Lib/EncoderLib/SEIFilmGrainAnalyzer.cpp
kjg0724: Done, moved it into the constructor next to calcVar/calcMean.

source/Lib/EncoderLib/SEIFilmGrainAnalyzer.h
kjg0724: Done, moved it to the public section.

K-os (approved):

PR #714, opened 2026-06-29
Arm: Fix Windows compilation with Clang and MSYS2

georges-arm:
The existing Windows build with Clang and MSYS2 fails in a couple of ways:

* The environment is not `UNIX OR MINGW` so the build currently fails to pick up the SVE and SVE2 flags. Change this to just `NOT MSVC`.

* `NVM_COMPILEDBY` is defined twice, causing compilation to fail on the duplicate `#define`. Fix the `#elif` chain so it only gets defined once.

* `bit_scan_reverse` was undefined since Clang appears not to define `__GNUC__` when on Windows, so add `|| defined( __clang__ )` to cover it.

* Delete the `CMAKE_SYSTEM_PROCESSOR` guard since this function is only called when `VVDEC_TARGET_ARCH` already matches.

### DIFF
--- a/cmake/modules/vvencCompilerSupport.cmake
+++ b/cmake/modules/vvencCompilerSupport.cmake
@@ -134,11 +134,7 @@ endfunction()
 # Check if the compiler supports the AArch64 SVE and SVE2 extensions, and set
 # variables for flags used to enable them to avoid duplication.
 function( set_if_compiler_supports_arm_extensions output_flag_sve output_flag_sve2 )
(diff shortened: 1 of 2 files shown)


georges-arm:
For context: https://www.msys2.org/ is the environment I've been tending to run Windows builds in since it is more Linux-shaped than the usual MSVC/MinGW alternatives and enables us to run Clang with the usual `-march=...` flags for feature enablement.

PR #715, opened 2026-06-30
Arm: Add Neon and SVE implementations for getPreBlkStatsAccum

athulya-arm:
- Add Neon and SVE implementation for getPreBlkStatsAccum.

Neon implementation improves performance by approximately 25% compared to the SIMDe version when benchmarked on a Neoverse V2 with LLVM 22.

SVE implementation improves performance by approximately 35% compared to the Neon version when benchmarked on a Neoverse V2 with LLVM 22.

- Add range-scaled tolerance support to unit tests

Add dynamic tolerance support to the unit-test comparison helpers using
minimum, relative, and maximum tolerance bounds. Apply it to weighted
`getPreBlkStatsWeightedAccum` checks, so small values remain bounded
by a floor while larger accumulated values get proportional tolerance.

### DIFF
--- a/source/Lib/CommonLib/arm/InitARM.cpp
+++ b/source/Lib/CommonLib/arm/InitARM.cpp
@@ -85,6 +85,15 @@ void AlfCovariance::initAlfCovarianceARM()
     _initAlfCovarianceARM<NEON>();
   }
 }
+
+void EncAdaptiveLoopFilter::initEncAdaptiveLoopFilterARM()
+{
+  auto vext = read_arm_extension_flags();
+  if( vext >= NEON )
+  {
+    _initEncAdaptiveLoopFilterARM<NEON>();
+  }
+}
 #endif
 
 #if ENABLE_SIMD_OPT_AFFINE_ME
--- a/source/Lib/CommonLib/arm/InitARM.cpp
+++ b/source/Lib/CommonLib/arm/InitARM.cpp
@@ -93,6 +93,12 @@ void EncAdaptiveLoopFilter::initEncAdaptiveLoopFilterARM()
   {
     _initEncAdaptiveLoopFilterARM<NEON>();
   }
+#if TARGET_SIMD_ARM_SVE
+  if( vext >= SVE )
+  {
+    _initEncAdaptiveLoopFilterARM<SVE>();
+  }
+#endif
 }
 #endif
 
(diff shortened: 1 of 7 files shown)


athulya-arm:
Hi @adamjw24 

We have a unit test failure for EncAdaptiveLoopFilter::getPreBlkStatsWeightedAccum with this patch. The unit test was merged recently from my side.

This function currently does not have a Neon or SVE implementation, and it involves floating-point operations. SIMD implementations can change the order of floating-point operations compared to the scalar reference, which may introduce very small numerical differences while still being mathematically correct. Because of this, we already added a tolerance of 0.1f in the unit test.

We believe this failure may have occurred due to the accumulated floating-point error randomly exceeding the current tolerance of 0.1f.

Would you be able to help us better understand the expected floating-point accuracy differences for this algorithm and advise on what an appropriate tolerance should be? We can either increase the tolerance more or swap to a relative tolerance. A relative tolerance might be preferred since the current failure occurs for large values (e.g. ~1e+6).

athulya-arm:
We now introduced a dynamic tolerance support to the unit-test comparison helpers with minimum, relative, and maximum tolerance bounds. We applied it to the weighted `getPreBlkStatsWeightedAccum` checks so that small values are bounded by a minimum tolerance, while larger values use a proportional (relative) tolerance.
For the different output buffer comparisons, the minimum tolerance values are set to 0.1f and 0.2f, and the corresponding maximum tolerance cutoffs are 1.0f and 2.0f.

Could you please take a look and let us know if these tolerance levels are acceptable with respect to the nature of the algorithm?

adamjw24:
> Hi @adamjw24
> 
> We have a unit test failure for EncAdaptiveLoopFilter::getPreBlkStatsWeightedAccum with this patch. The unit test was merged recently from my side.
> 
> This function currently does not have a Neon or SVE implementation, and it involves floating-point operations. SIMD implementations can change the order of floating-point operations compared to the scalar reference, which may introduce very small numerical differences while still being mathematically correct. Because of this, we already added a tolerance of 0.1f in the unit test.
> 
> We believe this failure may have occurred due to the accumulated floating-point error randomly exceeding the current tolerance of 0.1f.
> 
> Would you be able to help us better understand the expected floating-point accuracy differences for this algorithm and advise on what an appropriate tolerance should be? We can either increase the tolerance more or swap to a relative tolerance. A relative tolerance might be preferred since the current failure occurs for large values (e.g. ~1e+6).

Hi, sorry for the late answer. So this function might not be relevant soon, since we are in the process of preparing a removal of LMCS (its been depracated in the last version), which is the only tool needing it. For that reason larger tolerances are ok here.

adamjw24:
> We now introduced a dynamic tolerance support to the unit-test comparison helpers with minimum, relative, and maximum tolerance bounds. We applied it to the weighted `getPreBlkStatsWeightedAccum` checks so that small values are bounded by a minimum tolerance, while larger values use a proportional (relative) tolerance. For the different output buffer comparisons, the minimum tolerance values are set to 0.1f and 0.2f, and the corresponding maximum tolerance cutoffs are 1.0f and 2.0f.
> 
> Could you please take a look and let us know if these tolerance levels are acceptable with respect to the nature of the algorithm?

Looks good to me.

athulya-arm:
Thank you

PR #718, opened 2026-07-06
Arm: Add SVE dot-product implementation for MCTF calcVar

kjg0724:
Follow-up to #713, where @georges-arm suggested trying the SVE dot-product instructions for the variance computation in calcVar.

## Change
Adds `calcVar_sve`, an SVE variant of `calcVar_neon`. The variance pass replaces the Neon `vmull_s16` (lo/hi) + `vpadalq_s32` sequence with a single `vvenc_sdotq_s16(acc, pix, pix)` per group of eight int16 differences, which squares and accumulates into two int64 lanes in one instruction and widens to int64 natively, so the separate widening step is gone. The mean pass is unchanged.

It is wired into `MCTF::_initMCTF_ARM<SVE>()`. The film grain analyzer keeps using the Neon variant via the NEON init tier.

## Correctness
The result is bit-identical to `calcVar_neon` (same integer products, just a different accumulator partitioning that sums to the same total), so the existing exact-compare MCTF calcVar unit test covers it.

## Performance
Measured on a Neoverse V2, the variance kernel speeds up relative to the Neon version from about 1.2x on an 8x8 block to about 2x on a 64x64 block, with the gain growing for larger blocks.

## Testing
Tested on a Neoverse V2 with SVE enabled: the MCTF unit test (including the calcVar check) passes, and the benchmark numbers above are from the same machine. Note the SVE path is only exercised on SVE-capable hardware, so a green run on a non-SVE CI job does not cover it.

### DIFF
--- a/source/Lib/CommonLib/arm/sve/MCTF_sve.cpp
+++ b/source/Lib/CommonLib/arm/sve/MCTF_sve.cpp
@@ -49,6 +49,7 @@ POSSIBILITY OF SUCH DAMAGE.
 //  ====================================================================================================================
 #include "CommonDefARM.h"
 #include "CommonLib/CommonDef.h"
+#include "CommonLib/Rom.h"
 
 #include "MCTF.h"
 
@@ -62,6 +63,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #include "MCTF_neon.h"
 #include "mem_neon.h"
 #include "neon_sve_bridge.h"
+#include "../neon/sum_neon.h"
 #include <arm_neon.h>
 #include <arm_sve.h>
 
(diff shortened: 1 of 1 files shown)


kjg0724:
@georges-arm this implements the SVE dot-product suggestion from your review on #713. Tested on a Neoverse V2, unit tests pass, roughly 1.2x-2x speedup depending on block size. Let me know if you get a chance to take a look.

georges-arm:
> @georges-arm this implements the SVE dot-product suggestion from your review on #713. Tested on a Neoverse V2, unit tests pass, roughly 1.2x-2x speedup depending on block size. Let me know if you get a chance to take a look.

Hi @kjg0724. Thanks for following up with the SVE implementation. The code looks good to me.

kjg0724:
Thanks for taking a look, glad it works out.

PR #722, opened 2026-08-13
Arm: Add Neon implementation for Morph dilation

kjg0724:
Follow-up to the ARM SIMD porting stream from #713/#717/#718, as called out in #711 (Morph::dilation has no Neon implementation).

## Change
Adds an ARM Neon port of `Morph::dilation`. The scalar/x86 algorithm dilates a "strong pixel" mask: for each pixel, if any of its 3x3 neighbours (including itself) equals a sentinel `Value`, the output becomes `Value`, otherwise it keeps the original value.

Since `extendBorderPel(1,1)` already writes real replicated border pixels on all sides, including corners, before the pixel loop (same as the scalar reference), the Neon kernel does plain unaligned loads for the left/right/top/bottom taps instead of the in-register lane-shift-and-carry approach the x86 SIMD version uses to avoid touching the border. This keeps the kernel considerably simpler than the x86 version while producing the same result as the scalar reference.

Processes 8 pixels per vector, with an overlapping tail strip for widths that are not a multiple of eight (safe since the destination buffer is disjoint from the source within an iteration) and a scalar fallback for widths under eight.

`Morph`'s constructor now takes an `enableOpt` flag (default true), mirroring the pattern already used for `FGAnalyzer` in #713/#717, so the unit test can construct a scalar-only reference instance alongside an optimized one. `FGAnalyzer` forwards its own `enableOpt` to its embedded `Morph` member.

## x86 fix (second commit)
This PR is the first unit test coverage `Morph::dilation` has ever had, and it surfaced a pre-existing bug in the x86 SIMD path (`dilation_SIMD` in `FGAX86.h`): the vectorized loop bound `x < width - 8` underflows for `width < 8` (width is unsigned), causing out-of-bounds SIMD loads, and the loop plus its tail-column write assume width is a multiple of eight and strictly greater than sixteen (at exactly 16 the AVX2 branch's own main loop runs zero times), so other widths either produce silently wrong output (width == 8 or 16) or write past the destination buffer (e.g. width == 44). Added a width check that delegates to the scalar reference (`dilation_core`) for widths that don't meet those assumptions, instead of duplicating its 3x3 test or touching the vectorized code.

## Performance (third commit)
The first version of the Neon kernel used a column-outer traversal with a rolling row mask to cut compare instructions, but that meant each 8-pixel-wide strip re-streamed the whole frame height, which falls out of cache well before 4K. Switched to row-outer traversal (fresh row-mask computations per output vector, no rolling state) instead.

Measured on this machine (Apple M-series, `-O3`):

| resolution | scalar | Neon |
|---|---|---|
| 1920x1080, iter=4 | 36.70 ms | 2.92 ms (12.6x) |
| 3840x2160, iter=4 | 99.17 ms | 12.11 ms (8.2x) |

Both resolutions scale consistently; the column-outer version measured during review dropped noticeably from 1080p to 4K for the same comparison.

## x86 fix, height == 1 (fourth commit)
The same width guard from the second commit missed another pre-existing edge case: the row-pointer setup in `dilation_SIMD` assumes at least two rows, so at `height == 1` the `y == 0` branch makes `p_buf_down` point one stride past the single row, an out-of-bounds read that produces silently wrong output. Extended the same guard to also fall back to `dilation_core` when `height == 1`.

## Testing
`MorphDilation` unit test compares scalar vs optimized output exactly across widths {4,8,16,24,32,44,64} (covering the scalar-fallback path, the exact-vector-width and AVX2-boundary sizes, the overlapping-tail path, and the width needed to reach the vectorized path at height 1) and heights {1,4,8,33,44,64} (including the height == 1 edge case from the fourth commit), with `numIter` in {1,2,3,4} and `Value` in {strongPel, 0}, matching the call-site patterns in `findMask`. Passes on both ARM and x86 (the x86 build exercises the fixes in the second and fourth commits), and the full unit test suite has no regressions on either.

### DIFF
--- a/source/Lib/CommonLib/arm/neon/FGA_neon.cpp
+++ b/source/Lib/CommonLib/arm/neon/FGA_neon.cpp
@@ -102,7 +102,7 @@ template void FGAnalyzer::_initFGAnalyzerARM<NEON>();
 // never alias at real call sites. The two buffers generally have different
 // strides (buff has a margin, Wbuf usually has none), so they are indexed
 // with their own strides.
-static int dilationNeon( PelStorage* buff, PelStorage* Wbuf, uint32_t bitDepth, ComponentID compID, int numIter,
+static int dilation_neon( PelStorage* buff, PelStorage* Wbuf, uint32_t bitDepth, ComponentID compID, int numIter,
                          int iter, Pel Value )
 {
   CHECKD( buff == Wbuf, "buff and Wbuf must not alias" );
@@ -189,7 +189,7 @@ static int dilationNeon( PelStorage* buff, PelStorage* Wbuf, uint32_t bitDepth, 
 template <ARM_VEXT vext>
 void Morph::_initFGAMorphARM()
 {
-  dilation = dilationNeon;
+  dilation = dilation_neon;
 }
 
 template void Morph::_initFGAMorphARM<NEON>();
--- a/source/Lib/CommonLib/arm/neon/FGA_neon.cpp
+++ b/source/Lib/CommonLib/arm/neon/FGA_neon.cpp
@@ -96,6 +96,104 @@ void FGAnalyzer::_initFGAnalyzerARM()
 
 template void FGAnalyzer::_initFGAnalyzerARM<NEON>();
 
+// Morphological dilation: a pixel becomes Value if any of its 3x3 neighbours
+// (including itself) equals Value, otherwise it keeps its original value.
+// buff carries the (border-extended) input, Wbuf receives the output; they
+// never alias at real call sites. The two buffers generally have different
+// strides (buff has a margin, Wbuf usually has none), so they are indexed
+// with their own strides.
+static int dilationNeon( PelStorage* buff, PelStorage* Wbuf, uint32_t bitDepth, ComponentID compID, int numIter,
+                         int iter, Pel Value )
+{
+  CHECKD( buff == Wbuf, "buff and Wbuf must not alias" );
+
+  for( ; iter != numIter; ++iter )
+  {
+    Wbuf->bufs[0].copyFrom( buff->get( compID ) );
+    buff->get( compID ).extendBorderPel( 1, 1 );
+
+    const int       width  = buff->get( compID ).width;
+    const int       height = buff->get( compID ).height;
+    const Pel*      src    = buff->get( compID ).buf;
+    const ptrdiff_t sStr   = buff->get( compID ).stride;
+    Pel*            dst    = Wbuf->bufs[0].buf;
+    const ptrdiff_t dStr   = Wbuf->bufs[0].stride;
+
+    if( width >= 8 )
+    {
+      const int16x8_t vVal = vdupq_n_s16( Value );
+
+      auto rowMask = [vVal]( const Pel* row ) -> uint16x8_t {
+        int16x8_t  l = vld1q_s16( row - 1 );
+        int16x8_t  c = vld1q_s16( row );
+        int16x8_t  r = vld1q_s16( row + 1 );
+        uint16x8_t m = vceqq_s16( c, vVal );
+        m            = vorrq_u16( m, vceqq_s16( l, vVal ) );
+        m            = vorrq_u16( m, vceqq_s16( r, vVal ) );
+        return m;
+      };
+
+      for( int x = 0; ; x += 8 )
+      {
+        if( x + 8 > width ) x = width - 8;   // overlapping tail strip; safe, dst != src within an iteration
+        const Pel* p = src + x;
+
+        uint16x8_t mTop = rowMask( p - sStr );
+        uint16x8_t mMid = rowMask( p );
+        int16x8_t  cMid = vld1q_s16( p );
+
+        for( int y = 0; y < height; y++ )
+        {
+          uint16x8_t mBot    = rowMask( p + ( (ptrdiff_t)y + 1 ) * sStr );
+          uint16x8_t strong  = vorrq_u16( mTop, vorrq_u16( mMid, mBot ) );
+          int16x8_t  res     = vbslq_s16( strong, vVal, cMid );
+          vst1q_s16( dst + (ptrdiff_t)y * dStr + x, res );
+
+          mTop = mMid;
+          mMid = mBot;
+          cMid = vld1q_s16( p + ( (ptrdiff_t)y + 1 ) * sStr );
+        }
+
+        if( x + 8 >= width ) break;
+      }
+    }
+    else
+    {
+      // scalar fallback for width < 8, same 3x3 test as dilation_core
+      for( int y = 0; y < height; y++ )
+      {
+        for( int x = 0; x < width; x++ )
+        {
+          bool strong = false;
+          for( int dy = -1; dy <= 1 && !strong; dy++ )
+          {
+            for( int dx = -1; dx <= 1; dx++ )
+            {
+              if( buff->get( compID ).at( x + dx, y + dy ) == Value )
+              {
+                strong = true;
+                break;
+              }
+            }
+          }
+          dst[y * dStr + x] = strong ? Value : src[y * sStr + x];
+        }
+      }
+    }
+
+    buff->get( compID ).copyFrom( Wbuf->bufs[0] );
+  }
+  return iter;
+}
+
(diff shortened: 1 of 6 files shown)


source/Lib/CommonLib/arm/neon/FGA_neon.cpp
```diff
@@ -96,6 +96,104 @@ void FGAnalyzer::_initFGAnalyzerARM()
...
+  return iter;
+}
+
+template <ARM_VEXT vext>
+void Morph::_initFGAMorphARM()
+{
+  dilation = dilationNeon;
+}
+
+template void Morph::_initFGAMorphARM<NEON>();
```
georges-arm: No need for a separate instantiation here, just specialise for `NEON` directly:

```
template<>
void Morph::_initFGAMorphARM<NEON>()
{
  ...
}
```

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
```diff
@@ -96,6 +96,104 @@ void FGAnalyzer::_initFGAnalyzerARM()
 
 template void FGAnalyzer::_initFGAnalyzerARM<NEON>();
```
georges-arm: Move the other `init*` functions down to the bottom of the file so they are together? No need for a separate template instantiation here either.

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
```diff
@@ -96,6 +96,104 @@ void FGAnalyzer::_initFGAnalyzerARM()
...
+          uint16x8_t strong  = vorrq_u16( mTop, vorrq_u16( mMid, mBot ) );
+          int16x8_t  res     = vbslq_s16( strong, vVal, cMid );
+          vst1q_s16( dst + (ptrdiff_t)y * dStr + x, res );
+
+          mTop = mMid;
+          mMid = mBot;
+          cMid = vld1q_s16( p + ( (ptrdiff_t)y + 1 ) * sStr );
+        }
+
+        if( x + 8 >= width ) break;
```
georges-arm: Could this just be `x < width` in the loop condition?

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
```diff
@@ -96,6 +96,104 @@ void FGAnalyzer::_initFGAnalyzerARM()
...
+        for( int y = 0; y < height; y++ )
+        {
+          uint16x8_t mBot    = rowMask( p + ( (ptrdiff_t)y + 1 ) * sStr );
+          uint16x8_t strong  = vorrq_u16( mTop, vorrq_u16( mMid, mBot ) );
+          int16x8_t  res     = vbslq_s16( strong, vVal, cMid );
+          vst1q_s16( dst + (ptrdiff_t)y * dStr + x, res );
+
+          mTop = mMid;
+          mMid = mBot;
+          cMid = vld1q_s16( p + ( (ptrdiff_t)y + 1 ) * sStr );
```
georges-arm: Is there a need for this to be loop-carried? Seems like it would be cleaner code to just load it as `cMid = vld1q_s16( p + (ptrdiff_t)y * sStr );` at the start of each `y` iteration?

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
```diff
@@ -96,6 +96,104 @@ void FGAnalyzer::_initFGAnalyzerARM()
...
+    const Pel*      src    = buff->get( compID ).buf;
+    const ptrdiff_t sStr   = buff->get( compID ).stride;
+    Pel*            dst    = Wbuf->bufs[0].buf;
+    const ptrdiff_t dStr   = Wbuf->bufs[0].stride;
+
+    if( width >= 8 )
+    {
+      const int16x8_t vVal = vdupq_n_s16( Value );
+
+      auto rowMask = [vVal]( const Pel* row ) -> uint16x8_t {
```
georges-arm: I think there is potentially a faster expression to do this since currently the compare results are mostly redundant. In particular two of the compare result vectors could be reused across `y` loop iterations I think so it may be worth inlining this and seeing what you can simplify.

At the very least you can avoid the `c` load with something like:

```
        int16x8_t  l = vld1q_s16( row - 1 );
        int16x8_t  r = vld1q_s16( row + 1 );
        int16x8_t  lm = vceqq_s16( l, vVal );
        int16x8_t  rm = vceqq_s16( r, vVal );
        uint16x8_t m = vqtbl2q_u8( {lm, rm}, /*some indices to extract c from lm/rm*/ );
        return vorrq_u16( m, vorrq_u16( lm, rm ) );
```

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
kjg0724: Done, both init functions now use explicit specialisation.

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
kjg0724: Done, moved both to the end of the file next to each other.

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
kjg0724: Yes, done, that's cleaner. Verified by hand that clamping x to width-8 before the loop body still lets the plain x<width condition terminate at the right point on the next increment.

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
kjg0724: This whole strip-outer traversal is gone now. Benchmarked it against a row-outer version and the strip-outer one loses noticeably at 4K, since each strip re-streams the full frame height. Switched to row-outer, so there's no loop-carried state across y at all anymore.

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
kjg0724: Applied the c-load part: the centre row now computes its mask inline and reuses that load as the blend source instead of going through rowMask() twice for the same pixels.

Didn't do the cross-y reuse though. That needs consecutive y iterations for the same strip to be adjacent, which is exactly the traversal order I moved away from in the last commit (strip-outer was costing real throughput at 4K from re-streaming the frame per strip). Keeping reuse under row-outer would mean caching masks per strip across rows, which is more state for a saving that didn't show up in the benchmark even at 1080p where the frame fits in cache. Happy to revisit if you think it's worth it anyway.

kjg0724:
Found one more edge case while extending test coverage: `dilation_SIMD` also has an out-of-bounds read at `height == 1`, since the `y == 0` branch makes `p_buf_down` point one stride past the single row. Same fallback pattern as the width guard above. Pushed in 9b9f4c2, with height 1 and width 32 added to the unit test to cover it.

source/Lib/CommonLib/arm/neon/FGA_neon.cpp
georges-arm: Thanks for the explanation, this seems fine now.

georges-arm (commented):
Neon code seems fine now. There might be more tricks we can play by unrolling and using `EXT` to e.g. construct the cmp result for `p + 1` by using `EXT` to mix `p` and `p + 8` (from the next iteration of the loop) or `p - 1` by mixing `p - 8` (from the previous iteration of the loop) and `p`, but that is probably awkward to code and not guaranteed to be faster.

kjg0724:
Good idea in principle, but the buffer only has a 2-pixel margin (`margin = m_convWidthG / 2` in `SEIFilmGrainAnalyzer.cpp`), so an EXT-based carry can't pull a full 8-lane lookback/lookahead at the row edges without reading out of bounds. The first and last iteration of each row would still need the plain-load path, same as how the x86 version special-cases its own x==0 boundary. And since NEON's unaligned loads don't carry a real throughput penalty to begin with, the actual win here is just trading load-port pressure for permute-port pressure, which is genuinely uncertain without measuring on the specific core. I'll leave it as a follow-up rather than fold it into this PR, but happy to revisit if it turns out to matter somewhere.

adamjw24:
Thanks for providing the patch and the review!

PR #723, opened 2026-08-19
Arm: Add Neon implementation for Quant::NeedRdoq

kjg0724:
Follow-up to the ARM SIMD porting stream from #713/#717/#718/#722, as called out in #711 (Quant has no Neon implementation). Begins with `Quant::NeedRdoqSIMD`, the smallest of the three functions in `QuantX86.h` (`NeedRdoqSIMD`, `DeQuantCoreSIMD`, `QuantCoreSIMD`); the other two are follow-ups.

## Change
`Quant::xNeedRdoq` decides whether a transform block needs the RDOQ (rate-distortion optimized quantization) path at all: for each coefficient, it computes `abs(coeff) * quantCoeff`, adds a rounding offset, shifts, and returns as soon as any coefficient survives that shift as non-zero (otherwise false once the whole block is scanned).

The scalar survivor test `(abs(iLevel)*quantCoeff + offset) >> shift != 0` is rearranged into a per-call threshold: since the sum is always non-negative in the call domain, the shifted result is non-zero iff `abs(iLevel)*quantCoeff >= 2^shift - offset`, and since `quantCoeff` is always a positive constant this reduces to two signed compares, `iLevel >= levelThresh || iLevel <= -levelThresh`, with `levelThresh` computed once per call via a single integer division. The Neon kernel processes 8 coefficients per iteration (two `int32x4_t` compares per 4-lane chunk, OR-reduced across lanes via a pairwise-max/lane-read helper), with a 4-wide remainder step and a scalar tail for anything smaller. It returns as soon as a chunk's reduction is non-zero, mirroring the scalar reference's early exit. Only ARMv7-available intrinsics are used (no AArch64-only instructions, with a `vpmaxq_u32`/`vorr_u32` split for the lane reduction), since upstream still builds and CI-checks 32-bit Arm.

`Quant`'s constructor now takes an `enableOpt` flag (default true), mirroring the pattern already used for `FGAnalyzer`/`Morph` in #713/#717/#722, so the unit test can construct a scalar-only reference instance alongside an optimized one. `xNeedRdoq` moved from `private` to `public` for that purpose, following the same precedent (`DepQuant::m_checkAllRdCostsOdd1` etc.).

## Testing
`Quant` unit test compares scalar vs optimized output exactly, covering:
- a random sweep across `numCoeff` (multiples of four, matching real transform areas), the real `g_quantScales` table, and `shift` in [13,29] (traced from `getTransformShift()` + `iQBits`; an 8- or 10-bit 4x4 block at QP 63 reaches 29)
- directed cases: all-zero, a single non-zero coefficient at chunk boundaries (both extreme values, both signs), all-coefficients-at-minimum, and threshold-precision cases on both sides of the survival boundary for both offset rounding constants (171 and 256) and both signs
- a non-multiple-of-4 `numCoeff` case exercising the scalar remainder path, which real call sites never reach (`efArea` is always a product of two powers of two) but which is reachable through this public function pointer

Also verified bit-exact against scalar via an actual encode with selective RDOQ forced on (`--SelectiveRDOQ 1`), diffing the resulting bitstream MD5 against an unmodified build. Passes on both ARM and x86 (SIMDe), and the full unit test suite has no regressions on either; also verified against a 32-bit Arm build (compiles cleanly, no AArch64-only intrinsics).

## Performance
Microbenchmark on Apple Silicon (clang -O3), comparing scalar vs the current threshold-compare Neon kernel across representative transform-block sizes (4 to 2048 coefficients) and two input distributions (a block that never survives quantization, forcing a full scan, and one that survives on the first coefficient, exiting immediately):

Neon is faster across every size and distribution measured, 1.4x-9.3x depending on size/distribution (median ~7.2x over 3 repeated runs); restricted to the observed efArea range (8 and up, see below), the range is 2.2x-9.3x.

The threshold-compare kernel was also benchmarked against the original widening-multiply/64-bit-shift version on the same machine: roughly tied at 4 (below the observed efArea range), 10-21% faster at 8-16, and 1.5x-1.8x faster at 64-2048, consistent across full-scan and early-exit distributions over 3 repeated runs. (efArea=4 was not observed in practice: instrumented at the `xNeedRDOQ` call site, it did not occur across three encodes with `--SelectiveRDOQ 1` forced on, with an observed minimum of 8 (the 2x4 chroma case). The kernel still handles numCoeff=4 correctly, and the microbenchmark above shows it does not regress there.)

### DIFF
--- a/source/Lib/CommonLib/arm/neon/Quant_neon.cpp
+++ b/source/Lib/CommonLib/arm/neon/Quant_neon.cpp
@@ -137,13 +137,11 @@ static bool needRdoqNeon( const TCoeff* pCoeff, size_t numCoeff, int quantCoeff,
   return false;
 }
 
-template <ARM_VEXT vext>
-void Quant::_initQuantARM()
+template<>
+void Quant::_initQuantARM<NEON>()
 {
   xNeedRdoq = needRdoqNeon;
 }
-
-template void Quant::_initQuantARM<NEON>();
 
 }  // namespace vvenc
 
--- a/source/Lib/CommonLib/arm/neon/Quant_neon.cpp
+++ b/source/Lib/CommonLib/arm/neon/Quant_neon.cpp
@@ -61,13 +61,54 @@ POSSIBILITY OF SUCH DAMAGE.
 namespace vvenc
 {
 
-// Dispatch-only placeholder for M0: scalar-equivalent, to be replaced by
-// the real Neon implementation in M1.
+// Mirrors needRdoqCore (Quant.cpp): a reduction over the coefficient
+// buffer that returns true as soon as any coefficient survives
+// quantization. Real call sites always pass a numCoeff that is a
+// multiple of four (efArea = width * min(height,32), both powers of
+// two), so the bulk of the buffer is handled 8 coefficients at a time
+// with a 4-wide remainder step; a smaller remainder falls through to a
+// plain scalar loop.
+//
+// abs(coeff)*quantCoeff + offset is always non-negative in the real call
+// domain (coeff is clipped to +-2^15, quantCoeff and offset are always
+// positive), so vshlq_s64's arithmetic right shift and a logical shift
+// would produce identical bits here.
 static bool needRdoqNeon( const TCoeff* pCoeff, size_t numCoeff, int quantCoeff, int64_t offset, int shift )
 {
-  for( int uiBlockPos = 0; uiBlockPos < (int)numCoeff; uiBlockPos++ )
+  const int32x2_t vqnt   = vdup_n_s32( quantCoeff );
+  const int64x2_t voff   = vdupq_n_s64( offset );
+  const int64x2_t vshift = vdupq_n_s64( -(int64_t)shift );  // negative count -> right shift (SSHL)
+
+  auto lane = [&]( int32x2_t c ) -> int64x2_t {
+    return vshlq_s64( vaddq_s64( vmull_s32( c, vqnt ), voff ), vshift );
+  };
+
+  size_t i = 0;
+  for( ; i + 8 <= numCoeff; i += 8 )
   {
-    const TCoeff  iLevel   = pCoeff[uiBlockPos];
+    const int32x4_t c0 = vabsq_s32( vld1q_s32( pCoeff + i ) );
+    const int32x4_t c1 = vabsq_s32( vld1q_s32( pCoeff + i + 4 ) );
+
+    const int64x2_t l0 = lane( vget_low_s32( c0 ) );
+    const int64x2_t l1 = lane( vget_high_s32( c0 ) );
+    const int64x2_t l2 = lane( vget_low_s32( c1 ) );
+    const int64x2_t l3 = lane( vget_high_s32( c1 ) );
+
+    const uint64x2_t any = vreinterpretq_u64_s64( vorrq_s64( vorrq_s64( l0, l1 ), vorrq_s64( l2, l3 ) ) );
+    if( vget_lane_u64( vorr_u64( vget_low_u64( any ), vget_high_u64( any ) ), 0 ) != 0 )
+      return true;
+  }
+  if( i + 4 <= numCoeff )
+  {
+    const int32x4_t c0 = vabsq_s32( vld1q_s32( pCoeff + i ) );
+    const uint64x2_t any = vreinterpretq_u64_s64( vorrq_s64( lane( vget_low_s32( c0 ) ), lane( vget_high_s32( c0 ) ) ) );
+    if( vget_lane_u64( vorr_u64( vget_low_u64( any ), vget_high_u64( any ) ), 0 ) != 0 )
+      return true;
+    i += 4;
+  }
+  for( ; i < numCoeff; i++ )  // defensive: real call sites never leave a <4 remainder here
+  {
+    const TCoeff  iLevel   = pCoeff[i];
     const int64_t tmpLevel = ( int64_t ) std::abs( iLevel ) * quantCoeff;
     if( TCoeff( ( tmpLevel + offset ) >> shift ) != 0 )
       return true;
--- a/source/Lib/CommonLib/arm/neon/Quant_neon.cpp
+++ b/source/Lib/CommonLib/arm/neon/Quant_neon.cpp
@@ -41,7 +41,7 @@ POSSIBILITY OF SUCH DAMAGE.
 ------------------------------------------------------------------------------------------- */
 /**
  * \file Quant_neon.cpp
- * \brief Neon implementation of quantization functions for AArch64.
+ * \brief Neon implementation of quantization functions for Arm.
  */
 // ====================================================================================================================
 // Includes
--- a/source/Lib/CommonLib/arm/neon/Quant_neon.cpp
+++ b/source/Lib/CommonLib/arm/neon/Quant_neon.cpp
@@ -1,0 +1,90 @@
+/* -----------------------------------------------------------------------------
+The copyright in this software is being made available under the Clear BSD
+License, included below. No patent rights, trademark rights and/or
+other Intellectual Property Rights other than the copyrights concerning
+the Software are granted under this license.
+
+The Clear BSD License
+
+Copyright (c) 2019-2026, Fraunhofer-Gesellschaft zur Förderung der angewandten Forschung e.V. & The VVenC Authors.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification,
+are permitted (subject to the limitations in the disclaimer below) provided that
+the following conditions are met:
+
+     * Redistributions of source code must retain the above copyright notice,
+     this list of conditions and the following disclaimer.
+
+     * Redistributions in binary form must reproduce the above copyright
+     notice, this list of conditions and the following disclaimer in the
+     documentation and/or other materials provided with the distribution.
+
+     * Neither the name of the copyright holder nor the names of its
+     contributors may be used to endorse or promote products derived from this
+     software without specific prior written permission.
+
+NO EXPRESS OR IMPLIED LICENSES TO ANY PARTY'S PATENT RIGHTS ARE GRANTED BY
+THIS LICENSE. THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND
+CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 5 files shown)


source/Lib/CommonLib/arm/neon/Quant_neon.cpp
```diff
@@ -0,0 +1,131 @@
...
+  {
+    const TCoeff  iLevel   = pCoeff[i];
+    const int64_t tmpLevel = ( int64_t ) std::abs( iLevel ) * quantCoeff;
+    if( TCoeff( ( tmpLevel + offset ) >> shift ) != 0 )
+      return true;
+  }
+  return false;
+}
+
+template <ARM_VEXT vext>
```
georges-arm: No need for a separate specialisation here, just do something like...

```
template<>
void Quant::_initQuantARM<NEON>()
```

source/Lib/CommonLib/arm/neon/Quant_neon.cpp
```diff
@@ -0,0 +1,131 @@
...
+    const int32x4_t c0 = vabsq_s32( vld1q_s32( pCoeff + i ) );
+    const int32x4_t c1 = vabsq_s32( vld1q_s32( pCoeff + i + 4 ) );
+
+    const int64x2_t l0 = lane( vget_low_s32( c0 ) );
+    const int64x2_t l1 = lane( vget_high_s32( c0 ) );
+    const int64x2_t l2 = lane( vget_low_s32( c1 ) );
+    const int64x2_t l3 = lane( vget_high_s32( c1 ) );
+
+    const uint64x2_t any = vreinterpretq_u64_s64( vorrq_s64( vorrq_s64( l0, l1 ), vorrq_s64( l2, l3 ) ) );
+    if( vget_lane_u64( vorr_u64( vget_low_u64( any ), vget_high_u64( any ) ), 0 ) != 0 )
```
georges-arm: Since you only care about whether there exists a non-zero bit here you can use something else like `UMAXP` instead. There is no Neon `UMAXP` on `u64` but that doesn't matter here, can just cast it, e.g. https://godbolt.org/z/5oKP7fTz6 .

Note you'll need to add a helper here to avoid `vpmaxq_u32` being an AArch64-only instruction. See `sum_neon.h` for some existing workarounds (e.g. the `pairwise_add` ones) that you can copy.

source/Lib/CommonLib/arm/neon/Quant_neon.cpp
```diff
@@ -0,0 +1,131 @@
...
+      return true;
+  }
+  if( i + 4 <= numCoeff )
+  {
+    const int32x4_t c0 = vabsq_s32( vld1q_s32( pCoeff + i ) );
+    const uint64x2_t any = vreinterpretq_u64_s64( vorrq_s64( lane( vget_low_s32( c0 ) ), lane( vget_high_s32( c0 ) ) ) );
```
georges-arm: Simplify this line a bit by moving `l0` and `l1` out into separate variables to match the previous loop's layout?

source/Lib/CommonLib/arm/neon/Quant_neon.cpp
```diff
@@ -0,0 +1,131 @@
...
+// positive), so vshlq_s64's arithmetic right shift and a logical shift
+// would produce identical bits here.
+static bool needRdoqNeon( const TCoeff* pCoeff, size_t numCoeff, int quantCoeff, int64_t offset, int shift )
+{
+  const int32x2_t vqnt   = vdup_n_s32( quantCoeff );
+  const int64x2_t voff   = vdupq_n_s64( offset );
+  const int64x2_t vshift = vdupq_n_s64( -(int64_t)shift );  // negative count -> right shift (SSHL)
+
+  auto lane = [&]( int32x2_t c ) -> int64x2_t {
+    return vshlq_s64( vaddq_s64( vmull_s32( c, vqnt ), voff ), vshift );
```
georges-arm: Use `vmlal_s32` to combine the `mull` and `add`? The compiler might already be doing this for you.

source/Lib/CommonLib/arm/neon/Quant_neon.cpp
```diff
@@ -0,0 +1,131 @@
...
+    const uint64x2_t any = vreinterpretq_u64_s64( vorrq_s64( lane( vget_low_s32( c0 ) ), lane( vget_high_s32( c0 ) ) ) );
+    if( vget_lane_u64( vorr_u64( vget_low_u64( any ), vget_high_u64( any ) ), 0 ) != 0 )
+      return true;
+    i += 4;
+  }
+  for( ; i < numCoeff; i++ )  // defensive: real call sites never leave a <4 remainder here
+  {
+    const TCoeff  iLevel   = pCoeff[i];
+    const int64_t tmpLevel = ( int64_t ) std::abs( iLevel ) * quantCoeff;
+    if( TCoeff( ( tmpLevel + offset ) >> shift ) != 0 )
```
georges-arm, line 134: I think you should be able to calculate some threshold such that `threshold >> shift == 0` you can shortcut the test:

```
abs(iLevel) * quantCoeff + offset < threshold || abs(iLevel) * quantCoeff + offset > threshold
```

And since offset is constant, this can be rearranged as:

```
abs(iLevel) * quantCoeff < threshold - offset || abs(iLevel) * quantCoeff > threshold - offset
```

And I think this means you don't need the `abs` either:

```
iLevel * quantCoeff < threshold - offset || iLevel * quantCoeff > threshold - offset
```

Maybe you can also get rid of the multiply in this case, not sure but either way this may vectorise more efficiently?

kjg0724: Done. levelThresh = ceil((2^shift - offset)/quantCoeff), reducing the vector path to iLevel >= levelThresh || iLevel <= -levelThresh -- the multiply, abs, and 64-bit lanes are gone. quantCoeff is positive for every g_quantScales entry, and levelThresh maxes out at 34,783 in the real domain, well within s32. Directed tests cover the m-1/m boundary at both quantCoeff table extremes, the real minimum/midpoint/maximum shift values, and both offset factors; combinations whose boundary falls outside the real +-2^15 range are skipped. 10-21% faster at 8-16 coefficients and 1.5x-1.8x faster at 64-2048 than the previous kernel.

source/Lib/CommonLib/arm/neon/Quant_neon.cpp
kjg0724: Done, in adapted form. After the threshold rework (other thread) the reduction input is a u32 compare mask rather than s64 values, so this became a local any_lane_set_u32x4() helper: vpmaxq_u32 + a single u64 lane read on AArch64 (same idiom as your Godbolt), and vorr_u32 of the two halves on armv7, following the sum_neon.h pairwise_add pattern for the guard.

source/Lib/CommonLib/arm/neon/Quant_neon.cpp
kjg0724: The threshold rework in the other thread removed both the widening multiply and the offset add (offset is folded into the precomputed threshold), so there's nothing left to fuse here.

source/Lib/CommonLib/arm/neon/Quant_neon.cpp
kjg0724: Done via the threshold rework. Both loops now go through the same survivors() step with named mask variables, so the remainder block's layout matches the main loop.

source/Lib/CommonLib/arm/neon/Quant_neon.cpp
kjg0724: Done.

adamjw24:
Please mind that only the reviewer should be resolving the comments. Please do not resolve comments of other people.

I'd wait for the reviewers confirmation that all the aspects have been properly addressed.

kjg0724:
Understood, thank you for the note. I've reopened the 5 review conversations -- will wait for @georges-arm to confirm and resolve them.

source/Lib/CommonLib/arm/neon/Quant_neon.cpp
```diff
@@ -0,0 +1,150 @@
...
+      return true;
+  }
+  if( i + 4 <= numCoeff )
+  {
+    const uint32x4_t s0 = survivors( vld1q_s32( pCoeff + i ) );
+    if( any_lane_set_u32x4( s0 ) )
+      return true;
+    i += 4;
+  }
+  for( ; i < numCoeff; i++ )  // defensive: real call sites never leave a <4 remainder here
```
georges-arm: If this never happens then just `CHECKD` to assert it instead?

source/Lib/CommonLib/arm/neon/Quant_neon.cpp
georges-arm: Looks good, thanks!

georges-arm:
> Please mind that only the reviewer should be resolving the comments. Please do not resolve comments of other people.
> 
> I'd wait for the reviewers confirmation that all the aspects have been properly addressed.

Just FYI I don't think I have permission to resolve threads, but my old comments all seem resolved now so just the new `CHECKD` comment outstanding.

Thanks!

source/Lib/CommonLib/arm/neon/Quant_neon.cpp
kjg0724: Done. Kept CHECKD as a debug-mode assertion of the real-domain precondition (efArea is always a multiple of four), but kept the bounds-safe i+4<=numCoeff check plus scalar tail rather than dropping it: xNeedRdoq is a public function pointer and CHECKD is a no-op in release builds, so removing the tail entirely would let a non-multiple-of-4 call read past the buffer in release. Verified with ASan that a tight non-multiple-of-4 buffer crashes without the tail and doesn't with it.

georges-arm: So this tail loop is now dead code and can be removed?

kjg0724: Yes. Removed the tail and the numCoeff=3/7/13 test cases: they were exercising an input the CHECKD already rejects, so they'd throw in a Debug build anyway. In-domain coverage (4/8/12/16/2048 and the multiple-of-4 random sweep) is unchanged. This matches the other CHECKD kernels in arm/ (MCTF_sve, InterpolationFilter_neon), which also don't guard against out-of-contract sizes in release.

georges-arm: Looks good, thanks!

PR #725, opened 2026-09-10
Arm: Add Neon implementation for Quant::DeQuant

kjg0724:
Follow-up to #723, continuing the ARM SIMD porting stream tracked in #711. Second of the three functions in `QuantX86.h` (`NeedRdoqSIMD`, `DeQuantCoreSIMD`, `QuantCoreSIMD`); `QuantCoreSIMD` is the remaining follow-up.

## Change
`Quant::xDeQuant` dequantizes a full transform block: for each coefficient, it clips to `[-inputMaximum-1, inputMaximum]`, multiplies by `scale`, rounds and shifts by `rightShift` (a rounded right shift if positive, an unrounded left shift otherwise), and clips the result to `[-transformMaximum-1, transformMaximum]`.

A single `vrshlq_s32` with a per-lane shift count of `-rightShift` reproduces both scalar branches exactly -- SRSHL's rounded-right-shift semantics for a negative count match the scalar's `(v + (1<<(rightShift-1))) >> rightShift`, and its plain-left-shift semantics for a non-negative count match the scalar's unrounded `<< -rightShift` -- so the kernel has no branch on `rightShift`'s sign at all. `TCoeffSig` is 16-bit and `TCoeff`/`Intermediate_Int` are 32-bit, so one widening multiply (`vmull_n_s16`) into 32-bit lanes is enough. Only ARMv7-available intrinsics are used, since upstream still builds and CI-checks 32-bit Arm.

Real call sites (`Quant::dequant`'s `!enableScalingLists` path) pass a width (`maxX+1`) of 1 (degenerate ISP blocks), 2 (narrow chroma or ISP luma sub-partitions), 4, or a multiple of eight (ordinary transform sizes up to 64) -- never 3, 5, 6, or 7. Each width dispatches to an exactly-sized tight load/store.

The directed width-1 test case exposed a pre-existing bug in `DeQuantCoreSIMD`: its `maxX==0` branch shares a vector load/store with `maxX==1`, reading 6 bytes and writing 12 bytes past what a tightly allocated width-1 row owns. Fixed in the first commit (confirmed with ASan: the old code aborts with a heap-buffer-overflow on a tight buffer, the fixed code doesn't; the encodes described below, compared coefficient-for-coefficient, were otherwise bit-identical before and after this fix -- only the out-of-bounds access is removed). The Neon kernel doesn't repeat the bug.

`xDeQuant` moved from `private` to `public`, mirroring `xNeedRdoq` from #723, so the unit test can construct a scalar-only reference `Quant` instance and call it directly.

## Testing
`Quant` unit test compares scalar vs optimized output exactly, covering:
- a random sweep independently sampling shape, scale, and `rightShift` from the sets each is reachable over (shapes: ordinary transforms, ISP-degenerate 1xN/Nx1 blocks, and narrow 2xN/Nx2 blocks including 2x2 -- an SBT-halved chroma residual TU at an 8x4/4x8 inter CU; `rightShift`: `[-10,7]`, the full range `getTransformShift()`/`IQUANT_SHIFT` can produce over that domain per `Quant.cpp`) -- a conservative cross-product covering the reachable domain rather than only the exact (shape, rightShift) pairs a real encode would jointly produce
- directed cases for every width-dispatch branch (1/2/4/8/16/32/64): all-zero, representable extremes both signs, alternating sign, an isolated last-lane value in the width>=8 branch's second store
- the input-clamp branch: `inputMaximum` is 32767 for `rightShift >= -9`, but real call sites do reach `inputMaximum == 16383` at the domain's `rightShift == -10` endpoint (the 2x2 SBT-chroma case above) -- a directed case exercises exactly that real combination, in addition to a synthetic smaller-`inputMaximum` case for the clamp branch's general behavior
- a padded-stride case (also function-domain only: `piQCfStride` always equals width at real call sites)
- directed width==1 cases at heights 3, 5, 6, and 7 (crossing the width==1 NEON path's batch-of-4-rows/scalar-tail boundary added below), each with a tight and a padded stride, using distinct small alternating-sign coefficients per row so a lane-to-row mixup can't hide behind saturation

Verified bit-exact against scalar via encodes with `--DepQuant 0` (forces the ordinary dequant path; confirmed with a call counter that this hits `xDeQuant` ~343k times, all non-transform-skip) and `--TransformSkip 1` (forces transform-skip blocks to occur; confirmed ~21k `xDeQuant` calls, all transform-skip), at both 8- and 10-bit. Full unit test suite passed on both ARM and x86 (SIMDe), including under ASan with tightly allocated buffers for every width, and against a 32-bit Arm build -- against the kernel as of the first two commits.

Re-run against the current head (including the width==1/clamp restructuring below and the added boundary tests) under `-fsanitize=address,undefined` with the same tightly allocated buffers: the suite completes with no reported ASan/UBSan finding tracing to `Quant_neon.cpp` or `Quant.cpp`; a handful of pre-existing UBSan diagnostics unrelated to this PR remain elsewhere in the codebase (a generic test helper template, and NEON kernels in `MCTF_neon.h`, `InterPred_neon.cpp`, `AffineGradientSearch_neon.cpp`, `InterpolationFilter_neon.cpp`).

## Performance
Microbenchmark on Apple Silicon (clang -O3): a fixed-parameter kernel benchmark, not an end-to-end encoder measurement. Each shape times 200k calls with a fixed `scale=64`, `rightShift=2`, `inputMaximum=transformMaximum=32767`, a tight input stride, and reused input/output buffers, against the generic scalar `DeQuantCore` reference (not a width-1-specialized scalar loop) -- called through volatile function pointers (matching the real dispatch pattern -- `xDeQuant` is a function-pointer member, never inlined at the real call site). Table values are the median of 9 per-trial timing ratios, rerun 3x to check variance; updated after addressing review feedback below (loop structure, input clamping):

| shape | scalar/new speedup |
|---|---|
| 4x4 | ~5.1-5.2x |
| 8x8 | ~1.3x |
| 16x16 | ~1.3x |
| 32x32 | ~1.4x |
| 64x64 | ~1.4x |
| 1x16 | ~4.1-4.2x |
| 1x64 | ~5.6-5.7x |
| 2x8 | ~3.7-3.8x |
| 2x32 | ~3.4-4.5x across rerun medians (noisiest shape measured) |

The new kernel is faster than the generic scalar reference on every measured shape. Against the previous (pre-review-feedback) Neon version, 16x16-64x64 improved by roughly 1.24x-1.34x and 8x8/2x8 by roughly 1.05x-1.08x from the combined loop-structure-plus-clamping restructuring below (the benchmark doesn't isolate either change's individual contribution); width==1 (1x16, 1x64) showed the largest relative old-to-new improvement, roughly 1.9x-2.4x, from batching four rows per multiply/shift/store instead of one.

### DIFF
--- a/source/Lib/CommonLib/arm/neon/Quant_neon.cpp
+++ b/source/Lib/CommonLib/arm/neon/Quant_neon.cpp
@@ -146,15 +146,18 @@ static bool needRdoqNeon( const TCoeff* pCoeff, size_t numCoeff, int quantCoeff,
 // replaces DeQuantCore's Intermediate_Int arithmetic; no 64-bit lanes are
 // needed. Overflow check for the real (bitDepth, QP, shape) domain of this
 // codebase: the most negative rightShift this encoder's parameter formulas
-// can produce is -10, reached by an SBT-quartered 2x2 chroma sub-block (a
-// vertical-half SBT on an 8-wide inter CU, in 4:2:0, halved once more by a
-// second SBT split) at QP_per's maximum (10 at 8-bit, 12 at 10-bit). At that
-// shift inputMaximum -- derived at the call site as
+// can produce is -10, reached by an SBT-halved 2x2 chroma sub-block (e.g.
+// an 8x4 -- or 4x8 -- inter CU in 4:2:0 has 4x2 -- or 2x4 -- chroma, and a
+// single vertical- or horizontal-half SBT split halves that once more to
+// 2x2; SBT cannot recurse into a second split of an already-SBT-split TU,
+// so this is the only way to reach it) at QP_per's maximum (10 at 8-bit, 12
+// at 10-bit). At that shift inputMaximum -- derived at the call site as
 // (1 << (min(16, 25+rightShift) - 1)) - 1 -- is itself only 16383, not
-// 32767, so |q*scale| <= 16383*102 < 2^21, and the largest left shift (10)
-// keeps the shifted product under 2^31 - 1. inputMaximum only reaches its
-// ordinary 32767 for rightShift >= -9, where the same left-shift bound holds
-// a fortiori.
+// 32767, and clip(q, -inputMaximum-1, inputMaximum) admits the asymmetric
+// minimum -16384, so |q*scale| <= 16384*102 < 2^21, and the largest left
+// shift (10) keeps the shifted product under 2^31 - 1. inputMaximum only
+// reaches its ordinary 32767 for rightShift >= -9, where the same argument
+// (|q| <= 32768) holds a fortiori.
 //
 // Real call sites (Quant::dequant's !enableScalingLists path) always pass
 // a width (maxX+1) that is 1 (degenerate ISP), 2 (e.g. 2x4/4x2/2x2 chroma),
--- a/source/Lib/CommonLib/arm/neon/Quant_neon.cpp
+++ b/source/Lib/CommonLib/arm/neon/Quant_neon.cpp
@@ -134,17 +134,18 @@ static bool needRdoqNeon( const TCoeff* pCoeff, size_t numCoeff, int quantCoeff,
 // Mirrors DeQuantCore (Quant.cpp). Per coefficient:
 //   q  = clip(piQCoef[x], -inputMaximum-1, inputMaximum)
 //   v  = rightShift > 0 ? (q*scale + (1 << (rightShift-1))) >> rightShift
-//                        :  q*scale << -rightShift
+//                        :  q*scale * (1 << -rightShift)
 //   piCoef[x] = clip(v, -transformMaximum-1, transformMaximum)
 //
 // vrshlq_s32 with a per-lane shift count of -rightShift reproduces both
 // branches exactly: for rightShift>0 (negative count) SRSHL adds the same
 // rounding half-unit before an arithmetic right shift; for rightShift<=0
-// (non-negative count) it's a plain left shift, matching the scalar's
-// un-rounded `<< -rightShift`. TCoeffSig is int16, TCoeff/Intermediate_Int
-// are int32, so a single widening multiply (vmull_n_s16) into 32-bit lanes
-// replaces DeQuantCore's Intermediate_Int arithmetic; no 64-bit lanes are
-// needed. Overflow check for the real (bitDepth, QP, shape) domain of this
+// (non-negative count) it's a plain left shift, numerically the same as
+// the scalar's unrounded multiplication by `1 << -rightShift`. TCoeffSig is
+// int16, TCoeff/Intermediate_Int are int32, so a single widening multiply
+// (vmull_n_s16) into 32-bit lanes replaces DeQuantCore's Intermediate_Int
+// arithmetic; no 64-bit lanes are needed. Overflow check for the real
+// (bitDepth, QP, shape) domain of this
 // codebase: the most negative rightShift this encoder's parameter formulas
 // can produce is -10, reached by an SBT-halved 2x2 chroma sub-block (e.g.
 // an 8x4 -- or 4x8 -- inter CU in 4:2:0 has 4x2 -- or 2x4 -- chroma, and a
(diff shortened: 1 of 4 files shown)


source/Lib/CommonLib/arm/neon/Quant_neon.cpp
```diff
@@ -130,10 +131,90 @@ static bool needRdoqNeon( const TCoeff* pCoeff, size_t numCoeff, int quantCoeff,
...
+    int32x4_t v = vmull_n_s16( q, ( int16_t )scale );
+    v           = vrshlq_s32( v, vShift );
+    return vmaxq_s32( vminq_s32( v, vTMax ), vTMin );
+  };
+
+  if( width == 1 )
+  {
+    for( int y = 0; y <= maxY; y++ )
+    {
+      const int16x4_t q = vld1_lane_s16( piQCoef + y * piQCfStride, vdup_n_s16( 0 ), 0 );
```
georges-arm: Since the output is contiguous over `y` you could load multiple inputs up-front (i.e. unroll to process two iterations from `y` at a time and do a pair of `vld1_lane`s)?

For `width == 1` you might also want to check that this is actually faster than just implementing it as normal C scalar arithmetic?

source/Lib/CommonLib/arm/neon/Quant_neon.cpp
```diff
@@ -130,10 +131,90 @@ static bool needRdoqNeon( const TCoeff* pCoeff, size_t numCoeff, int quantCoeff,
...
+      vst1q_lane_s32( piCoef + y, convert( q ), 0 );
+    }
+    return;
+  }
```
georges-arm, line 242: Move the other `else if (width == N)` conditions out of the `for` loop and duplicate the loop for each of them to match the `width == 1` case?

source/Lib/CommonLib/arm/neon/Quant_neon.cpp
```diff
@@ -130,10 +131,90 @@ static bool needRdoqNeon( const TCoeff* pCoeff, size_t numCoeff, int quantCoeff,
...
+    else if( width == 4 )
+    {
+      vst1q_s32( dst, convert( vld1_s16( src ) ) );
+    }
+    else
+    {
+      for( int x = 0; x < width; x += 8 )
+      {
+        const int16x8_t s = vld1q_s16( src + x );
+        vst1q_s32( dst + x, convert( vget_low_s16( s ) ) );
```
georges-arm: You might want to move the input clamping out of the `convert` function since it is being done separately on low/high halves of the vector here but it could be done just once on the full vector.

source/Lib/CommonLib/arm/neon/Quant_neon.cpp
kjg0724: Done — width==1 now batches four rows using four `vld1_lane_s16` loads, one vector multiply/shift/output-clip sequence, and one `vst1q_s32`, since the output row stride is one `TCoeff`. Remaining rows use the existing single-lane NEON sequence. Against the generic scalar reference in this Apple Silicon/clang -O3 microbenchmark (`scale=64`, `rightShift=2`, reused buffers), 1x16 improved from about 2.2x to 4.1-4.2x over scalar, and 1x64 from about 2.4x to 5.6-5.7x. This didn't compare against a separately specialized width-1 scalar loop.

source/Lib/CommonLib/arm/neon/Quant_neon.cpp
kjg0724: Done — width dispatch now occurs outside the row loops. Widths 1, 2, and 4 each have their own loop and return; width>=8 uses the final loop.

source/Lib/CommonLib/arm/neon/Quant_neon.cpp
kjg0724: Done — the width>=8 path now clamps the full 8-lane load once with `vminq_s16`/`vmaxq_s16`, then splits it into low/high halves for the multiply/shift. The combined loop-and-clamp restructuring measured roughly 1.24x-1.34x over the previous version at 16x16/32x32/64x64 in this benchmark; that comparison doesn't isolate the clamp change's contribution.

georges-arm:
Neon code looks good now, thanks for the changes.

K-os:
> Neon code looks good now, thanks for the changes.

Thank you very much for your in-depth review.

